PR #39, opened 2023-01-10
Add email and search types to TextField

dimfeld:
These just translate to the `type` parameter on the `input`, and are otherwise treated as normal "text" types.


techniq:
Thanks @dimfeld.  This PR made me realize I still had Netlify configured, but it is no longer now.  I switched to Vercel a good while ago, but left them both running (which worked until I switched from the `static` to `vercel` adapter).

PR #40, opened 2023-01-10
Update instructions to explicitly say not to use tailwindcss-forms

dimfeld:
It adds borders around text fields which look really bad when used in a svelte-ux TextField. Seemed worth noting since it's a very common plugin to use.


techniq:
Good to know.  I've not used that plugin myself (occasionally use `tailwindcss-typography`).  Thanks for the PR

PR #41, opened 2023-01-11
replace all uses of hardcoded blue with accent color

dimfeld:
I looked through the doc pages with the accent color changed, and they seemed good. 

Not totally sure about ErrorNotification though. This changes the button text color to the accent color which I think is what I would want but I didn't see a page to test that component in the docs so I haven't actually looked at it.


techniq:
I don't I have a doc page for it yet, but the change looks good.   Thanks

PR #42, opened 2023-01-11
Add tailwind-merge and use it on AppBar

dimfeld:
Noticed that the `bg-white` wasn't taking effect on the AppBar "color" example so I added `twMerge`. Unlike at work I did not use `clsx` on it since twMerge already has its own clsx-like implementation internally.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "svelte-ux",
-  "version": "0.22.4",
+  "version": "0.23.0",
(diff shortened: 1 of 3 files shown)


techniq:
I decided to go [all in](https://github.com/techniq/svelte-ux/commit/e46fce30e4684d2ef95e0456aba34177a50fe7f1) on `tailwind-merge` (and added the same `cls` wrapper).  Released as `0.24.0`

PR #102, opened 2023-10-30
Fix typos in doc site

adamnationx:
Fixes a couple of typos I noticed on [svelte-ux.techniq.dev](https://svelte-ux.techniq.dev/).


techniq:
Thanks @adamnationx!

PR #103, opened 2023-10-30
TableOfContents examples!

jpoles1:
A table of contents example, which I think covers the functionality of the component. Going to work on tackling TreeList next.


techniq:
Thanks so much @jpoles1!  I really appreciate it.

PR #109, opened 2023-11-04
Replace fieldsets with divs.

### LINKED ISSUE #108: Prevent forced use of <fieldset> for individual fields.
I noticed while using the package that TextField and SelectField both have a `<fieldset>` at the root while most other inputs have a `<div>`. This causes problems for me in a codebase I am attempting to use this package for since I need to be able to put multiple inputs/fields into one fieldset (and also include a `<legend>` as a header for that subset of fields).

Would it be possible to update those two (and any others I missed) to likewise swap `<fieldset>` for `<div>` so that users can handle that aspect themselves? Or if you don't want to break compatibility, perhaps use `<svelte:element this={rootElement}>` with an `export let rootElement = "fieldset"`?

I'd be happy to submit a PR myself if approved.

willnationsdev:
Closes #108. There are now no more mentions of "fieldset" within the codebase except for a TODO in `ToggleGroup.svelte`.

Note that I had a large number of warnings/errors in my VS Code project, but everything built (and seemed to behave) just fine. May have just misconfigured something on my end, idk.

1. Gave the new `<div>` elements a "group" role.
2. Identified each element/component supporting `disabled` and made sure to pass the parent component's state to it since the `<fieldset>` previously would've affected nested form fields' `disabled` state automatically.
    - The `Input` component did not have a `disabled` property, so I had to add one.

If I missed anything, please feel free to let me know (or just fix w/ a follow-up).

### DIFF
--- a/src/lib/components/Field.svelte
+++ b/src/lib/components/Field.svelte
@@ -55,8 +55,8 @@
   let labelEl: HTMLLabelElement | null = null;
 </script>
 
-<fieldset
-  {disabled}
+<div
+  role="group"
   class={cls(
     'Field',
     'group flex gap-1',
@@ -165,6 +165,7 @@
             {#if clearable && hasValue}
               <Button
                 icon={mdiClose}
+                {disabled}
                 class="text-black/50 p-1"
                 on:click={() => {
                   value = Array.isArray(value) ? [] : typeof value === 'string' ? '' : null;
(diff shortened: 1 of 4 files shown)


techniq:
@willnationsdev Thanks!  Now that you mention it, I think `disabled` was maybe another reason I used `fieldset`.  In some cases, I'll add `pointer-events-none` on the container to make sure nothing is clickable in the disabled state, but I like passing `disabled` down in this case, especially since that's how it worked (inherently).

I gave it a run through and everything looked unchanged.  Thanks again!

PR #110, opened 2023-11-04
Typehint & A11y ignore for Button.

willnationsdev:
Strict TS hints in VS Code display errors/warnings without these adjustments.

The ignore hint is necessary because we can't dictate the library user's correct application of ARIA roles and because we cannot conditionally add the `on:click` forwarding in response to the value of `href`.

### DIFF
--- a/src/lib/components/Button.svelte
+++ b/src/lib/components/Button.svelte
@@ -31,7 +31,7 @@
     | 'text'
     | 'none'
     | undefined = undefined; // default in reactive groupContext below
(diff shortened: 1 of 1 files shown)


techniq:
Thanks @willnationsdev.  There is definite room for improve with typescript (some of which are related to Svelte's current shortcomings with typing within the markup).

These PRs are great so feel free to keep them coming as you find things.  I've become a little immune to the console warnings at this point 😁

PR #115, opened 2023-11-08
Allow mixed null/undefined values in Actions<T>.

willnationsdev:
This removes some typing errors in TextField.svelte (finally figured out how to get the actions to not complain).

### DIFF
--- a/src/lib/actions/input.ts
+++ b/src/lib/actions/input.ts
@@ -3,7 +3,7 @@ import type { Action } from 'svelte/action';
 /**
  * Auto focus node when rendered.  Useful for inputs
  */
-export const autoFocus: Action<HTMLElement, { delay?: number }> = (node, options?) => {
+export function autoFocus(node: HTMLElement, options?: { delay?: number }) {
   // TODO: Add options to "restoreFocus" on destroy()
   // const elementFocused = document.activeElement as HTMLElement;
   setTimeout(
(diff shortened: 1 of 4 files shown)


src/lib/actions/input.ts
```diff
@@ -3,7 +3,7 @@ import type { Action } from 'svelte/action';
 /**
  * Auto focus node when rendered.  Useful for inputs
  */
-export const autoFocus: Action<HTMLElement, { delay?: number }> = (node, options?) => {
```
techniq, line 6: I assume using the Svelte provided `Action<>` interface is causing some of your problems?

techniq (approved):

src/lib/actions/input.ts
willnationsdev: Yeah, *something* about the way it's being interpreted with the current syntax causes it to assume misinformation about the method signature and I couldn't seem to figure out what I was doing wrong, but when I changed the style of the function declaration, the issue just *magically* disappeared, lol.

techniq: I feel like I've been getting more actions complaining in typescript lately... not sure if something changed in Svelte 4, Svelte UX, or I've just never noticed them as much.  If it works for you, I'm all for it :)

willnationsdev: It changed in Svelte 4. They made it so that it now requires a generic argument for the arguments you pass to the action when it's used.

PR #125, opened 2023-11-11
Fix type errors for Button's use of Icon data.

willnationsdev:


willnationsdev:
Just found something to improve with this, so don't merge it quite yet.

willnationsdev:
Okay, now the PR introduces inclusive input types for supplying icon data as arguments for components whether it be a null, undefined, string, IconDefinition, or an object with all the properties of an Icon component (which should include Icon components themselves). Then it extracts the `data` representation of the input and passes that along to the inner `Icon` component's `data` argument. It can therefore accommodate a variety of input types and empower ancestor components to seamlessly pass those arguments to descendants until they finally reach an `Icon`, and without introducing varying requirements on the data types of the passed data. Should be able to gradually update component arguments to use these new `IconInput` types if that sounds good to you.

willnationsdev:
Amended the changeset to provide more information due to the above changes.

packages/svelte-ux/src/lib/utils/icons.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type IconInput = ComponentProps<Icon>['data'] | ComponentProps<Icon>;
+export type IconData = ComponentProps<Icon>['data'];
+
+export function asIconData(v: IconInput): IconData {
+  return isIconComponentProps(v) ? v.data : v;
+}
+
+function isIconComponentProps(v: any): v is ComponentProps<Icon> {
+	return typeof(v) === "object" && v && typeof(v['classes']) === "object" && "path" in v;
```
techniq, line 13: There is a [isLiteralObject()](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/utils/object.ts#L4C27-L6) util that could help refine this a little.

I'm also not sure `typeof(v['classes']) === "object" && "path" in v` is the correct check.  

I wonder if `asIconData()` should just be... 

```ts
export function asIconData(v: IconInput): IconData {
  return ('data' in v) ? v.data : v;
}
```

I might be overlooking something though.

.changeset/slimy-starfishes-report.md
```diff
@@ -0,0 +1,6 @@
+---
+'svelte-ux': patch
+---
+
+Added new `IconInput` and `IconData` types to enable inclusive & seamless passing of icon arguments between components. Also provides a `asIconData` utility function for type-safe conversion.
+Fixed type errors for Button & TextField's use of Icon data.
```
techniq, line 6: `<Icon>` used to just accept `<Icon path="...">`

...then it was extended to allow `<Icon svg="...">`, ``<Icon svgUrl="http://...">`

...then it was extended to have a simple `<Icon data="...">` prop and infer which of the props was intended (path, svg, svgUrl), so I call that a feature to allow `TextField` to take in more options than just `path` :).

`svgUrl` is quite useful when your loading over the network, and not populating your bundle with all potential options.

packages/svelte-ux/src/lib/utils/icons.ts
```diff
@@ -0,0 +1,14 @@
+
+import type { ComponentProps } from "svelte";
+import type { default as Icon } from "$lib/components/Icon.svelte";
+
+export type IconInput = ComponentProps<Icon>['data'] | ComponentProps<Icon>;
+export type IconData = ComponentProps<Icon>['data'];
+
+export function asIconData(v: IconInput): IconData {
```
techniq, line 8: What if we call this `getIconData()`?

techniq (approved):
This looks great, and a huge win.  Just a few minor notes

.changeset/slimy-starfishes-report.md
willnationsdev: Hehe, I wasn't saying that it doesn't support different kinds of inputs. More that, we have instances of wrappers with inner wrappers with inner-inner Icons, etc. and the allowed data types for those wrapper components' distinct arguments do not always match, leading to inconsistency that has the potential to force type errors for no reason. I actually have to compliment this project for the diversity of data formats that `Icon` supports.

techniq: :).  It's been battled tested in a lot of use cases.  One thing I struggle with is getting the time to show all the niceties in examples :).  `<Table>` is under represented for how much it packs ;)

packages/svelte-ux/src/lib/utils/icons.ts
willnationsdev: > There is a [isLiteralObject()](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/utils/object.ts#L4C27-L6) util that could help refine this a little.

I agree, that's better to use.

> I'm also not sure `typeof(v['classes']) === "object" && "path" in v` is the correct check.

When I was writing this initially, I had thought that `ComponentProps<Icon>` required an object with all the same properties of `Icon`, but I can see now that it actually is like a `Partial` wrapper around that, so yeah, this definitely would *NOT* be the right check. Good catch.

Since `IconDefinition` has a more consistent structure & because it's the only other supported "object" type, we should just check to see if the given `v` object does NOT have certain properties that all `IconDefinition` objects are required to have and that `Icon` lacks. Therefore, a single property check can, by process of elimination, validate which type we're dealing with.

> I wonder if asIconData() should just be...

In theory, you'd be right (if it were just JS), but in order for TypeScript to rule out the possibility that the `IconInput` parameter is actually a `ComponentProps<Icon>`, you either have to exhaustively check all possible properties (far more work than is actually necessary) or write a function that forcibly coerces the type understanding via a type guard return value (`<param> is <type>` return type).

packages/svelte-ux/src/lib/utils/icons.ts
willnationsdev: I considered that initially, and I'm good with that if you prefer it. Only reason I chose `as` was because `get` sounded like it was consistently extracting data, but just in potentially different/abstracted ways whereas `as` communicates that the object in question might already be the return type (`IconData`) and conveys that the function's purpose is ultimately type conversion rather than extraction. Up to you.

packages/svelte-ux/src/lib/utils/icons.ts
techniq: fwiw, there is a [typeGuards](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/types/typeGuards.ts) utils with `hasProperty()`, etc.  Might not be useful here

packages/svelte-ux/src/lib/utils/icons.ts
techniq: We can leave it `as` is (pun intended) :)

PR #127, opened 2023-11-12
Replace SelectList with SelectField.

willnationsdev:
- Updates `QuickSearch` to use `SelectField`.
- Removes `SelectList`.
- Adds features to `SelectField` that allow it to support `SelectList`'s use cases.


packages/svelte-ux/src/lib/components/QuickSearch.svelte
```diff
@@ -62,12 +64,15 @@
   }}
 >
   <div class="overflow-auto max-h-[min(90dvh,600px)] min-w-[400px] py-1">
-    <SelectList
+    <SelectField
       icon={mdiMagnify}
       placeholder="Search..."
+      showToggleIcon={false}
```
techniq, line 70: maybe invert this an call it `hideToggleIcon`.  I like boolean props to always default to `false` and not require a value.

packages/svelte-ux/src/lib/components/SelectField.svelte
```diff
@@ -352,10 +374,15 @@
     on:blur={onBlur}
     on:keydown={onKeyDown}
     on:keypress={onKeyPress}
-    actions={(node) => [selectOnFocus(node)]}
-    class={cls('h-full', theme.field, classes.field)}
+    actions={fieldActions}
+    classes={{
+      root: cls('h-full'),
```
techniq, line 379: `cls()` isn't needed since we are not doing any overrides with later parameters (or non-string values like objects and arrays)

packages/svelte-ux/src/lib/components/SelectField.svelte
```diff
@@ -49,13 +60,17 @@
...
+  $: fieldClasses = typeof(classes.field) === "string" ? { root: classes.field } : classes.field;
+
   // Menu props
   export let placement: Placement = 'bottom-start';
   export let autoPlacement = true;
   export let matchWidth = true;
   export let resize = true;
   export let disableTransition = false;
   export let menuProps: ComponentProps<Menu> | undefined = undefined;
+  export let optionsMode: "list" | "menu" = "menu";
```
techniq, line 73: Maybe call it `inlineOptions` since we can’t really support more than 2 modes (is always either `Menu` and passed props, or `undefined` which ignores the props)

packages/svelte-ux/src/lib/components/SelectField.svelte
```diff
@@ -352,10 +374,15 @@
     on:blur={onBlur}
     on:keydown={onKeyDown}
     on:keypress={onKeyPress}
-    actions={(node) => [selectOnFocus(node)]}
-    class={cls('h-full', theme.field, classes.field)}
+    actions={fieldActions}
+    classes={{
+      root: cls('h-full'),
+      ...theme.field,
+      ...fieldClasses,
```
techniq, line 381: since these are all objects, if anyone defines something (like `root`), it will remove all the other classes.  This might be OK, just thinking it through...

techniq (changes requested):
A few comments.  I also noticed an opening issue while testing it in the vercel preview I'll share is discrod

packages/svelte-ux/src/lib/components/SelectField.svelte
willnationsdev: Oh, true, it wouldn't be additive in that case. Not sure how much I like that...if only you could easily pipe each overlapping property through `cls()`...hmmm.

techniq: Could maybe use [merge()](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/utils/object.ts#L66-L79), although it uses lodash, and I've been trying to ween the project off of it (either have it's own impl in the project, or use [just](https://github.com/angus-c/just/), although I don't think it had a similar merge as lodash's `mergeWith`

techniq:
@willnationsdev I'm not sure if you would want to tackle this as part of the same PR, but I created #132 (extracted from #130) which helps standardize `SelectField` with `MenuField` (mostly small tasks)

packages/svelte-ux/src/lib/components/SelectField.svelte
willnationsdev: Reverted this tidbit.

packages/svelte-ux/src/lib/components/QuickSearch.svelte
willnationsdev: Made this change.

packages/svelte-ux/src/lib/components/SelectField.svelte
willnationsdev: Made this change.

packages/svelte-ux/src/lib/components/SelectField.svelte
willnationsdev: Reverted this change.

willnationsdev:
Was in the process of implementing changes to incorporate #132, but ran into #152 which is now a blocker that prevents me from fully testing & verifying that whatever changes I've made are actually working properly.

willnationsdev:
@techniq This should now be ready for review. I've included the requested changes from #132.

techniq:
Awesome, thanks @willnationsdev.  I don't have a ton of time to look at it right now, but will more this evening.

I did a quick run through using the Vercel [preview](https://svelte-ux-git-fork-willnationsdev-select-ui-techniq-team.vercel.app/docs/components/SelectField) and noticed QuickSearch has an issue.


https://github.com/techniq/svelte-ux/assets/177476/60df4a16-1fc5-443a-9f3c-d79939e136ff

My guess is we just need to update where we [build](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/routes/+layout.svelte#L46) the options to use `label` instead of `name`.

willnationsdev:
@techniq QuickSearch should be fixed now. I updated all the SelectField docs, but never updated the `label`/`name` difference in the `+layout.svelte` file, as you suspected.

techniq (changes requested):
Looks really good, and I love all the Typescript improvements.  Some 2 small nits (and I can handle them post merge if you want.

Could you add the `text-left` class to `<button>` within `SelectField` and make sure it doesn't cause any unexpected regressions, and maybe add a SelectField example with `inlineOptions` enabled?

packages/svelte-ux/src/lib/components/SelectField.svelte
```diff
@@ -331,12 +358,15 @@
...
   }
 </script>
 
 <!-- svelte-ignore a11y-click-events-have-key-events -->
-<div role="button" tabindex="-1" aria-pressed={open ? "true" : "false"} class={cls('SelectField', theme.root, classes.root, $$props.class)} on:click={onClick}>
+<button
+  aria-haspopup={!inlineOptions ? "listbox" : undefined}
+  class={cls('SelectField block w-full cursor-default', theme.root, classes.root, $$props.class)}
+  on:click={onClick}>
+
```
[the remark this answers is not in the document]
willnationsdev: `text-left` added & example created. I also moved the `options` class overrides from the QuickSearch into the `SelectField` and conditionally add them in if `inlineOptions={true}` is specified.

willnationsdev:
@techniq Afaik, these changes should be ready for merging. Please review it when you get a chance & let me know if I need to update anything else.

techniq (approved):
LGTM, Thanks!  We chatted about fixing the chevron button toggle, but we'll get to that at some point :).  I'm glad to get this merged so I can restack on top of it for the theme work, and this PR didn't have Cloudflare previews (assuming because it existed before I switched) and looking to shutdown my Vercel integration (save money and stop messaging me about failures because I switched to use the cloudflare sveltekit adapter explicitly).

Thanks again, and sorry it took so long to merge.  I'll try to update one of my large projects "soon" and see if I find any unexpected regressions.

Issue #129, opened 2023-11-13
[format] Improve internationalization 


techniq:
Need to update [format](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/utils/format.ts#L21), particularly the [formatNumberAsStyle](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/utils/number.ts#L53-L55) to make sure it handles other currencies (ex. `€` vs `$`, and possibly other separators (like commas))

See [discussion](https://discord.com/channels/920755200552226868/1172179335009017898)

techniq:
Implemented via https://github.com/techniq/svelte-ux/pull/134

PR #134, opened 2023-11-15
:construction: UPDATE: format & number #129

jycouet:
- [x] add tests for numbers
- [x] Add app defaults.
- [x] Add app presets


techniq:
@jycouet This is looking great.

> Do we need formatNumber & formatNumberAsStyle? Or one could be enough?

I would *love* to only have a single `formatNumber()` that handles all use cases.  I think renaming `formatNumberAsStyle()` to `formatNumber()` is likely the best approach.  I believe using `d3-format` gives us the most control of formatting compared to `Intl.NumberFormat`, although it could be I haven't explored it deep enough.  I've been meaning to add a test suite for this and appreciate you taking the time to do so.  I do have a basic format [doc](https://svelte-ux.techniq.dev/docs/utils/format) page to show some of the `format()` usage (which calls `formatNumberAsStyle()` for numbers).

A quick list of use cases off the top of my head

## Convenient formatting

Using simple string presets (`integer`, `decimal`, `currency`, etc) to conveniently format a number.  This is used by `Table`, `SpringValue`, and `TweenedValue` in [Svelte UX](https://github.com/search?q=repo%3Atechniq%2Fsvelte-ux%20formattype&type=code), and `Axis`, `Labels`, `Legend`, and `TooltipItem` in [LayerChart](https://github.com/search?q=repo%3Atechniq%2Flayerchart%20formattype&type=code) (maybe others).  Short/rounded `metric` values are nice, especially for Chart axis labels or summarized dashboard values (not the best name tbh, but the best I could come up with at the time.).

`format()` is basically a facade to format numbers along with dates (ex. `PeriodType.Day`) and applying adhoc functions.  Updating date formatting to support locales would likely be something to look at soon (based on your needs).  It's going to have similar issue with "convenient" presets but also nice to have full format string support for custom formatting, and might be even tricky in practice :(.


## Internationalization

Initially I had hopes to just use the browser's default locale (such as what `navigator.language`) but I know it's not that easy, and likely needs to be user supplied/override capable.  It looks like d3-format's [formatLocale](https://d3js.org/d3-format#formatLocale) / [formatDefaultLocale](https://d3js.org/d3-format#formatDefaultLocale) was the magic sauce we needed.

At first glance it looks like you need to be explicit with these formats, unlike `Intl.NumberFormat()` which takes in the `locale`.  Am I overlooking something, or would we need/want to create a map of locales to return the appropriate settings?


This looks great, and i'm on board with both simplifying and supporting non-English locales/Intl.

If it would be helpful to sync up on a video chat, let me know and I can try to free up some time.

jycouet:
The only drawback of `d3-format` compared to `Intl.NumberFormat` is the size. But it's probably ok.

Automatic local (`navigator.language`) is good and bad. In my case, I don't care about user browser settings, I want to display `€`.

Main point for me is setting once option, then make use of it. (now writing this, maybe it should be any array of options... and take the first one (or a default one) if not specified)

Great feedback 👍, Yes, let's have a video chat soon.

techniq:
> The only drawback of `d3-format` compared to `Intl.NumberFormat` is the size. But it's probably ok.

Agreed.  If we can get similar flexibility using `Intl.NumberFormat` I'm all for using it.  It's mostly supporting the presets ('integer', 'decimal', 'currency', 'percent', 'percentRound', and 'metric').  

Metric (significant digits) might be tricky.  It looks like `unit: byte` might work, although I'm not sure if this would be using `1024` or `1000` when "rounding".  I typically use this to show large summarized currency values (50K, 10M dollars, etc) and often for chart axis labels.

```js
const byteValueNumberFormatter = Intl.NumberFormat("en", {
  notation: "compact",
  style: "unit",
  unit: "byte",
  unitDisplay: "narrow",
});

console.log(byteValueNumberFormatter.format(10));
=> 10B

console.log(byteValueNumberFormatter.format(200000));
=> 200KB

console.log(byteValueNumberFormatter.format(50000000));
=> 50MB
```

If we think it's worth it (and I think it might), we can try to go through all the presets and find the applicable settings.  We might also be able to add a little pre/post processing to get them "just right" (for example take off the "B" (byte) for "metric".

I like the idea of dropping a dependency :).  [Intl.NumberFormat](https://caniuse.com/?search=intl.numberformat) has great browser support.


> Automatic local (`navigator.language`) is good and bad. In my case, I don't care about user browser settings, I want to display `€`.
> 
> Main point for me is setting once option, then make use of it. (now writing this, maybe it should be any array of options... and take the first one (or a default one) if not specified)

Yeah, I can of suspected that.

> Great feedback 👍, Yes, let's have a video chat soon.

Thanks, and sounds good.

jycouet:
All right @techniq I guess that you can have a look 👍 
I left a few comments / TODO / & fail test in purpose for you to comment

packages/svelte-ux/src/lib/utils/format.ts
```diff
@@ -1,24 +1,41 @@
...
+export function format(
+  value: string | Date | null | undefined,
+  format?: PeriodType,
+  ...extraFuncArgs: any[]
+): string;
+export function format(value: any, format?: FormatType, ...extraFuncArgs: any[]): any {
   let formattedValue = value ?? ''; // Do not render `null`
 
   if (format) {
+    // TODO QUESTION: is this used?
```
techniq, line 29: It is.  It's helpful for `<Table>` column formats when you want to do something special, such as call another format function like `humanizeDuration()` or any other formatting like only show the last X digits of an ID (SSN, SHA, etc)

```svelte
<Table columns={[
  // ...
  {
    name: 'startDate',
    format: PeriodType.Day,
  },
  {
    name: 'cost',
    format: 'currency'
  },
  {
    name: 'duration',
    format: value => humanizeDuration({ start: value });
  }
]} />
```

The same is useful in LayerChart to format Axis ticks (`<Axis format={value => ...} />`), and there are other examples.

packages/svelte-ux/src/lib/utils/number.test.ts
```diff
@@ -47,3 +47,165 @@ describe('round()', () => {
...
+
+  it('returns value as string for style "none"', () => {
+    const actual = formatNumber(1234.5678, { style: 'none' });
+    expect(actual).equal('1234.5678');
+  });
+
+  it('formats number with integer', () => {
+    const actual = formatNumber(1234.5678, { style: 'integer' });
+    expect(actual).equal('1234');
+    expect(actual).equal('1,234'); // TODO: Today its like this... It's intended? (I leave the test breaking for now)
```
techniq, line 70: It is intended.  I find it more readable to have integer/counts with commas... such a... `12,345 results` vs `12345 results` (or even bigger numbers).

Maybe there is a use case to not have this?  If so, this might be an example where we would want to be able to customize the `style` (presets) via context like we chatted about today.

packages/svelte-ux/src/lib/utils/number.ts
```diff
@@ -1,45 +1,109 @@
...
     return '';
   }
 
-  const formatter = Intl.NumberFormat(undefined, {
+  if (options.style === 'none') {
+    return `${number}`;
+  }
+
+  if (options.style === 'integer') {
+    return `${parseInt(number.toString())}`;
```
techniq, line 40: This should also round so `1234.56` becomes `1,235`

packages/svelte-ux/src/lib/utils/number.test.ts
techniq: What if we formatted integers using...

```js
Intl.NumberFormat('en-US', {
  maximumSignificantDigits: value?.toString().split('.')[0].length
})
```

see [REPL](https://svelte.dev/repl/bc26d92b40e942b79395889759f92fa1?version=4.2.4)

packages/svelte-ux/src/lib/utils/number.ts
jycouet: let's speak about all this on the previous comment, I revolve this one

packages/svelte-ux/src/lib/utils/number.test.ts
jycouet: - It's a US thing to have `,` for 1_234 (in europe this would be very confusing)
- the name "integer" almost implies an integer to me (no formats (,) & no rounding)

Actually, what is a `integer` today is: `formatNumber(1234.5678, { fractionDigits: 0 });` _(knowing that default style is decimal)_

I understand the need, just I'm not sure about the terme "integer"

packages/svelte-ux/src/lib/utils/number.test.ts
```diff
@@ -47,3 +47,165 @@ describe('round()', () => {
...
+    expect(actual).equal('200K dollars');
+  });
+
+  it('dollars 50M', () => {
+    const actual = formatNumber(50000000, {
+      style: 'metric',
+      suffix: ' dollar',
+    });
+    expect(actual).equal('50M dollars');
+  });
```
techniq, line 210: Maybe also include a non-suffix'd example (just returns `50M`)

packages/svelte-ux/src/lib/components/setup.ts
```diff
@@ -0,0 +1,35 @@
+import type { FormatNumberOptions } from '$lib/utils/number';
+import { getContext, setContext } from 'svelte';
+import type { Theme } from './theme';
+
+export type Setup = {
```
techniq, line 5: I think I'd like to call this `Settings` instead of `Setup` (everywhere, not just this instance).

I could see apply an override of settings somewhere in your component tree (since it's just context).  For example maybe if you had routing with the locale (`/en/...`, `/fr/...`, etc), or you wanted to style a page/section different (frontpage marketing vs app), etc.

It would also be nice to have a `<Settings>` component like `<Theme>` (which just makes it easier to set the context)

packages/svelte-ux/src/routes/customization/+page.md
```diff
@@ -45,15 +45,31 @@ Internally, each component uses the `cls()` util which leverages [tailwind-merge
...
+      container: 'hover:shadow-none group-focus-within:shadow-none', // same as <TextField classes={{ container: '...' }}>
+    },
+  },
+
+  formats: {
+    // This is the default, but you can override it here for your app globally
+    numbers: {
+      locales: 'en',
+      currency: 'USD',
+      fractionDigits: 2,
```
techniq, line 70: Is there any risk of having only a single numbers options, or benefit to having options per style/preset?

For example, if `fractionDigits: 4`, wouldn't all currency be set with 4 (since it doesn't specify a min/max in formatNumber()), or would you want to be able to specify them separately...

```js
{
  formats: {
    numbers: {
      default: {
        locales: 'en',
      },
      currency: {
        currency: 'USD',
        fractionDigits: 2,
      },
      decimal: {
        fractionDigits: 4
      }
    }
  }
}
```

Just kind of thinking out loud

packages/svelte-ux/src/lib/utils/number.ts
```diff
@@ -1,81 +1,93 @@
...
+  | 'percent' // from Intl.NumberFormat options.style NumberFormatOptions
+  | 'unit' // from Intl.NumberFormat options.style NumberFormatOptions
+  | 'none'
+  | 'integer'
+  | 'percentRound'
+  | 'metric';
+
+export type FormatNumberOptions = Intl.NumberFormatOptions & {
+  style?: FormatNumberStyle;
+  locales?: string | undefined;
```
techniq, line 15: I was thinking this should be called `locale` instead of `locales`, but the argument is called that.  TIL 😁

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat#using_locales

packages/svelte-ux/src/lib/utils/number.ts
```diff
@@ -1,81 +1,93 @@
...
+  fractionDigits?: number;
+};
 
 // See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat
 export function formatNumber(
   number: number | null | undefined,
-  options: Intl.NumberFormatOptions & { fractionDigits?: number } = {}
+  options: FormatNumberOptions & {
+    suffix?: string;
```
techniq, line 23: Why are `suffix` and `suffixExtraIfMany` not part of `FormatNumberOptions`?

techniq:
@jycouet This is really good!  A few more small comments, but I think that will be it.

Also, the Vercel preview build is failing due to `createTheme()` still be referenced in a handful of places.  I think we just need to replace `<Theme>` with `<Settings>` and update `createTheme({...})` with `settings({...})` in these locations.

image

image

packages/svelte-ux/src/lib/components/setup.ts
jycouet: All right, I'll update to `Settings`.

The override thing, you can already do it no?

packages/svelte-ux/src/routes/customization/+page.md
jycouet: Yeah, it's also an options, I like it. I'm hesitant... Adding `default` right now is a good thing.
Let's see during my next night 😅

packages/svelte-ux/src/lib/utils/number.ts
jycouet: hahaha, I did 20 go & back on this one, and then saw why it could be an array & fallback & co.
Smarter ppl did this, let's follow 😉

packages/svelte-ux/src/lib/utils/number.ts
jycouet: Heuuuuu, was late yesterday. Thx for looking 💪
😅

packages/svelte-ux/src/routes/customization/+page.md
jycouet: Okay... To be checked with what I just did

techniq:
@jycouet I'm ready to merge if you are.  I left a note about whether we should default to `en`/`USD`, but if you're happy with that, I am.

packages/svelte-ux/src/lib/components/settings.ts
```diff
@@ -0,0 +1,51 @@
...
+    return getContext<Settings>(settingsKey) ?? {};
+  } catch (error) {
+    return {};
+  }
+}
+
+export function getFormatNumberOptions(style?: FormatNumberStyle) {
+  let toRet = {
+    locales: 'en',
+    currency: 'USD',
```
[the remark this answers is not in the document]
jycouet: I would keep it like this as default.
Let's see when people have a dedicated use case.

With local I will put 3000€, locally everything will work and Co, but people in Hungary will see 3000 Forint (that's 8€)!!! 😅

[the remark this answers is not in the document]
techniq: Awe, very good point!  Thanks!

techniq (approved):
Awesome work!

Issue #136, opened 2023-11-20
Replace date-fns's formatDate with Intl.DateTimeFormat


techniq:
Similar to the recent work to use `Intl.NumberFormat` instead of `d3-format` for `formatNumber()`, it would be great use [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) instead of date-fns's [format](https://date-fns.org/v2.16.1/docs/format) for [formatDate](https://github.com/techniq/svelte-ux/blob/74099102133c0929dfba55ad6cf6bb8d1fe4e524/packages/svelte-ux/src/lib/utils/date.ts#L552)

- [x] Add internationalization support
- [x] Add `settings()` support
- [x] Support existing `PeriodType` and variant (short/long)
- [x] Evaluate existing date-fns format() usage (sometimes aliased as `dateFormat` or `formatDate`) and possibly replace.  Might be difficult as these usually allow passing date format [strings](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table))

One day we might be able to replace date-fns with [Temporal](https://tc39.es/proposal-temporal/docs/index.html).  There is a [polyfill](https://github.com/js-temporal/temporal-polyfill#readme) but not sure it's worth investigating yet.

After formatting, there are a lot of date components (`DateField`, `DatePickerField`, `DateRangeField`, ... ) that would need updated for Intl support (maybe we just allow a global date format string (ex. `MM/dd/yyyy`) in `settings()` that we use for these types of use cases.

techniq:
@jycouet An issue to track thoughts / discussions :)

PR #137, opened 2023-11-20
🕰️ NEW: formatDate

jycouet:
## DO NOT MERGE

Link to: https://github.com/techniq/svelte-ux/issues/136

- [x] add changeset (breaking rmv `dateDisplay`)
- [x] update `dateDisplay` Doc
  - [x] all relevant things move to `format`
  - [x] doc: `component/dateDisplay` to remove (use https://svelte-ux.techniq.dev/docs/components/dateDisplay/) to compare)
  - [x] remove `dateDisplay.ts`
- [x] update doc to explain format with date


packages/svelte-ux/src/lib/utils/date.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  it('should format date string', () => {
+    expect(formatDate(DATE, PeriodType.Day, 'short')).equal('11/21');
+    expect(formatDate(DATE, PeriodType.Day, 'long')).equal('Nov 21, 2023');
+  });
+
+  it('should format date for PeriodType.Week', () => {
+    expect(formatDate(DATE, PeriodType.WeekSun, 'short')).equal('11/21 - 11/27');
+    expect(formatDate(DATE, PeriodType.WeekSun, 'long')).equal('11/21/2023 - 11/27/2023');
+    // QUESTION: What is WeekSun supposed to be? why is it different from WeekMon? (I would expect the first date to start on Sunday or monday? or?)
```
techniq, line 30: Yes, this is to tied to the start of the period (week or bi-week).  Technically bi-week has 14 different starts (odd/even week, Sun - Sat).

In the past I've used this for different pay periods (some companies are Sun - Sat, some are Mon - Sun, etc) as well as other aggregations we calculated via a statistics engine.

UI wise this is controlled by...

image

Although it currently appears to be having an issue too.  I'll resolve this as I'm working on the double click preset issue.

Note, I never exposed all 14 bi-week starts in this UI as it was confusing to most users, and luckily the few customers shared the same start weeks.

Also, CalendarYear (Jan start) and FiscalYear (Oct) should really be a single year with configurable start month (Jan - Dec).  This is a bit of an implementation leak with how I used it in the past.   Fiscals typically start in October, January, or July in the states, but really they can start about anytime.

Not sure it worth refining ATM, though.

jycouet: But then: 
```ts
expect(formatDate('2023-11-21', PeriodType.WeekSun, 'short')).equal('11/21 - 11/27');
```
Should not pass no?
Shouldn't it gives `11/19 - 11/25`?

techniq: That’s right. I’m not sure why it’s not passing TBH other than it’s typically not used. 

Usually when I’ve used these periods it’s been with an explicit start/end date and would use the Day format for each end. 

I’ve passed these period types server side to express the type of group by / aggregation to perform (ex. For the last 6 months, roll up payroll hours bi-weekly)

jycouet: I like a lot the idea of these formats. (I keep doing things on the "spot", but having a global thinking is a good thing)
I'll have a look

techniq: Thanks.  They help a lot for both consistency, and not having to think too hard 😁.

Btw, we'll probably need to scan the repo and try to remove most/all explicit `date-fns`'s `format` within components, such as [Month](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Month.svelte#L94) and [DateRange](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/DateRange.svelte#L171).  Searching for `'date-fns'` to look at imports is probably the easiest way, since the function is called `format` like Svelte UX's, and its usage is typically aliased.

LayerCake has a lot of these as well.

We can iteratively do these, just letting you know in case you start using these components and wonder why they are not in the expected language/format 😁

packages/svelte-ux/src/lib/utils/date.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  });
+
+  it('in without offset == UTC', () => {
+    const localDate = '2023-11-21T04:00:00';
+    const utcDate = localToUtcDate(localDate);
+    expect(utcDate.toISOString()).equal('2023-11-21T04:00:00.000Z');
+  });
+});
+
+describe('getMonthDaysByWeek()', () => {
```
jycouet, line 98: Could you check if it's the behavior you expect? (It was strange before, so I refactored, and remove a dep!!! :) )

techniq: I love the idea of removing a `dep` :).  

I believe this is off by 1 and should return `2023-10-29` - `2023-12-02` instead of `2023-10-30` - `2023-12-03`, unless I'm not following the timezone correctly.

This is used on the [Month](https://svelte-ux.techniq.dev/docs/components/Month/) component, IIRC

image

jycouet: Perfecto!

You expect Sunday first day of the week in US? or monday?

techniq: Sunday :), but that would be locale specific.

jycouet: Dates... I love it! But actually, I might have realized something! Something BIG! (Or maybe I'm just stupid haha).
Could you pull the branch and do me a screenshot of your month component?
(I will explain after :p)

techniq:
Using the PR [preview](https://feat-formatdate.svelte-ux.pages.dev/docs/components/Month/), or did you need me to run the server on my machine?

image

packages/svelte-ux/src/lib/utils/date.test.ts
jycouet: Forget it 🤣, It should be good now.

jycouet:
Reading https://github.com/techniq/svelte-ux/issues/150 and with my current progress, I think that we have to do some choices now :).

For `yesterday`, `3 days ago`, ... this is working well:
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat

For date format, ... this is working well:
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

I started to add a `dico`, for things that I can't find in `intl` in a robust way.
https://github.com/techniq/svelte-ux/blob/feat/formatDate/packages/svelte-ux/src/lib/utils/date.ts#L97-L105

---

I think that we will not get rid of `date-fns` as there are so many good utilities and `intl` is only about formatting.
But, I think that formatting should be done in 1 single way either `intl` either `date-fns`, but not both (my current state!)

## intl
- It's cool as it's native js support (less to ship in the browser)
- It's cool as it's similar to number that we did before
- It's cool, with 1 string, it's supports so many locales!
- It's not cool as it will be a breaking change for the lib (https://svelte-ux.techniq.dev/docs/components/DateField format will have to be done another way)

## date-fns for formatting
- people will be forced to add locales of date-fns
- keeping the way it's done
- less work as it's spread in a lot of places

jycouet:
Check out: https://svelte-ux-git-feat-formatdate-techniq.vercel.app/docs/utils/format#playground

techniq:
> Reading #150 and with my current progress, I think that we have to do some choices now :).
> 
> For `yesterday`, `3 days ago`, ... this is working well: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat
> 
> For date format, ... this is working well: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat
> 
> I started to add a `dico`, for things that I can't find in `intl` in a robust way. https://github.com/techniq/svelte-ux/blob/feat/formatDate/packages/svelte-ux/src/lib/utils/date.ts#L97-L105
> 
> I think that we will not get rid of `date-fns` as there are so many good utilities and `intl` is only about formatting. But, I think that formatting should be done in 1 single way either `intl` either `date-fns`, but not both (my current state!)
> 
> ## intl
> * It's cool as it's native js support (less to ship in the browser)
> * It's cool as it's similar to number that we did before
> * It's cool, with 1 string, it's supports so many locales!
> * It's not cool as it will be a breaking change for the lib (https://svelte-ux.techniq.dev/docs/components/DateField format will have to be done another way)
> 
> ## date-fns for formatting
> * people will be forced to add locales of date-fns
> * keeping the way it's done
> * less work as it's spread in a lot of places
> * convenient for `DateField` format

I really like the idea of using `Intl` for all the reasons you mentioned, especially not having to manage/download locales (i.e. `date-fns/locale`) and the simplicity/consistency.  With that said, when I took a look at the PR preview playground, I was initial surprised by the DateField format until I realized that what one of the things you mentioned :).

image

I also think `date-fns` provides a lot of additional utility beyond formatting so I'm good with keeping it.

For `DateField` `format` default, I wonder if we just keep our own locale/format lookup for this use case.

```js
const defaultFormat = {
  en: 'MM/dd/yyyy`,
  fr: 'dd/MM/yyyy',
  ...
}
```
if we overlook one, we could have it just default to 'MM/dd/yyyy' and if a user complains, we'll know which one we missed :).

We could also take a date like `2023-11-21`, format it using [toLocaleDateString(locale)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString#using_locales), then parse where the values ended up, but this might be overkill.

I think the simple map might be the best, as for a default, just getting MDY or DMY or YMD right is good enough.  I think always using `/` as a divider is probably fine too, although I know some locales (or even personal preference) might want to use `-` or `.`.

Looking at https://en.wikipedia.org/wiki/Date_format_by_country and https://docs.oracle.com/cd/E19455-01/806-0169/overview-7/index.html, I think we could build up this list fairly easily.  Maybe even make it a `case` statement to simplify the impl (including handling the fallback default).

Lastly, I think we should expose this format on `settings()` for when a dev/user wants to override the format.

How does that sound?  We can try to sync up in a few days and chat about it further if you want too.

jycouet:
Oups, I just read your message after spending my all afternoon on it 🤣🤞
_But, I think we are thinking the same way, hahaha_

Check out the update ✅

_I love all tests, I'm ssssoooo confident moving things around. And if something is not tracked, it's easy to add it 👍_
I tried to clean as much as possible.

I think that I would like to remove `dateDisplay`! to really cut `format` from `date-fns`, but it might break some people code... So I don't know....

I'll do a break now to party a bit 🎉
Speak to you soon

techniq:
Happy New Year (not quite here yet)!  You've been busy 😅.

I love the idea of removing `dateDisplay` / consolidating into a single date format function for all use cases (or requiring dateFns format() directly when needed).  I know there is a little usage of `dateDisplay` within [LayerChart](https://github.com/search?q=repo%3Atechniq%2Flayerchart%20datedisplay&type=code).  There was even a [commit](https://github.com/techniq/layerchart/commit/aa0d4b049a882058d2cd947137fc4733ed335f3d) that switched from dateFns's `format()` to `dateDisplay(..., { format: ... })` to better handle null so it's not a 100% drop in if we require dateFns for this kind of formatting, but I'm not against it.  If you're planning to support custom format strings in `formatDate()` that would be awesome too (it could still use dateFns, but then I guess we're back with dealing with locales in this case).  We might just need to support more PeriodTypes for example time-only (`h:mm:ss`).  Full disclosure: `PeriodType` has/had a tight coupling with a backend aggregation solution in one of my projects.  I no longer work on that project, so I'm not as concerned anymore :).

Regarding the variants, currently we have `short`, and `long`.  I've [considered](https://github.com/techniq/svelte-ux/blob/fd53036a31d7f7477a9f81f00b763bfc924fd475/packages/svelte-ux/src/lib/utils/date.ts#L555) switching this to `sm`, `lg` and supporting `md`, `xxs`, `xl`, `2xl`, etc (similar to Tailwind breakpoints) as needed.  If we do switch to this variant naming, we should do the same for [duration](https://github.com/techniq/svelte-ux/blob/fd53036a31d7f7477a9f81f00b763bfc924fd475/packages/svelte-ux/src/lib/utils/duration.ts#L99) / [Duration](https://github.com/techniq/svelte-ux/blob/fd53036a31d7f7477a9f81f00b763bfc924fd475/packages/svelte-ux/src/lib/components/Duration.svelte#L12).

jycouet:
## breakpoint
I'm not a big big fan of `short` & `long`, I added `default` & `custom`, and now I think that it's "ok".
But I like even less `sm`, `lg`, `md`, ...

I think that it's hard to find 6 different levels of display of a `year` for example! (And then, at what breakpoint you go from `yy` to `yyyy`?) => I always forget levels & breakpoints... Ok you could customize... But then, it's too much of flexibility as of today. (My 2 cents)
`d`, `dd`, `eee`... All this is `Day`, what breakpoint is what?
I think that breakpoints don't fit, maybe I'm missing something?
Maybe you change my mind with some idea of "why"?

That's why I was thinking:
- Default => Ok, by default what do we want?
- Short => Let's have a good short version
- Long => Let's have a good long version
- Custom => to escape and do what you want

Often, Default is or `short` or `long`... (but not always the same).
I'm almost thinking: "do we even need short & long"?

## dateDisplay
> . If you're planning to support custom format strings in formatDate() that would be awesome too (it could still use dateFns, but then I guess we're back with dealing with locales in this case)

The custom, is already doing "a lot", and we can handle more and more cases if we have feedback.

I'll look into your commit and add different flags to help doing the same.

## side node
The `fns` format "MM d, yyyy" is equivalent to `intl` format "d MM yyyy", because my "transform" takes care only about tokens, and locales + intl will do the right thing for US => "MM d, yyyy" 😉
I'm pretty happy about this cool trick. (You can also keep "MM d, yyyy")
Just not sure how to document this.

techniq:
## Breakpoint

Good points, I think it would be too much customization (especially to support).  FWIW, it's not based on viewport/breakpoints, mostly usage based on available real estate / context.  For example, LayerChart axis labels typically use `short` for dates since a simple MM/YY is enough context (and space constrained), and you typically use a tooltip to get more details on the date (get year, day of week, etc).  This screenshot isn't the best example, but does show the concept a little.

image

So really `short` / `long` is based on usage, but I like `default` and `custom`.  I wonder if we need `custom` but allow you to change `short` and `long` (similar to the number presets of `decimal`, `currency`, etc) and maybe let `default` allow you to specify `short` and `long` only?  Trying to find the sweet spot of simplicity but not overly restrictive is a delicate balance.  Thoughts?

## dateDisplay

I need to get myself up to speed on `DateTimeFormat` and read through the PR closer, but it all sounds great.   It looks like you are mapping [unicode](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table) date tokens (aka date-fns format string) to DateTimeFormat options (i.e. 'long', 'short', '2-digit', 'numeric', `narrow`, `undefined`, ...).  Looking on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), I was having trouble finding the full list of available options for each date part (year, month, day, etc).  Did you find a good reference?

## side node

Kind of what I mentioned above, but really we're translating unicode format strings and mapping them to DateTimeFormat options.  Do this mean `MM/dd/yyyy` and `MM-dd-yyyy` and `yyyy-MM-dd` will also result in the same output string based on the locale, or can you still keep the explicit formatting?  I'm assuming no, which is probably OK (and using date-fns `format()` directly when needed is probably best for this).  How to document this (and if there is a better approach to make this more obvious if so).  Basically all that matters is each part's definition, and not the order or separators.

Side note: I'm really liking how this is coming together and I REALLY appreciate you digging in.  I knew it wouldn't be a small task, but this will be another huge improvement to Svelte UX.  Thanks for all the effort!

packages/svelte-ux/src/lib/utils/date.ts
```diff
@@ -549,11 +575,181 @@ export function formatISODate(
   return formatISO(date, { representation });
 }
 
+export enum DateToken {
```
jycouet, line 578: Highlight: Currently handle tokens. (you can always escape and use IntlOptions)

jycouet:
🥳🥳🥳 Youhouuu, I think I'm getting close to something I like (and hopefully you too 😉)

Trying to summarize:
You can configure presets for `days`, `weeks`, `months`, `monthsYears`, `years` with => `short`, `default` and `long` variants. These variants are either `tokens` either `intlOptions` (it's beautiful!)

Then, when you call `format`, you pass the `PeriodType` and the `variant`. (Or you can always pass `custom` and do what you want).

By default, everything is configured, but you can tweak a lot... I think that it's a quite nice balance spot.

I didn't find a great mapping with token, but did it case by case. And now that we can anyway pass all `intlOption`, I think that it's a sweet spot too.

I'll remove the draft more for you too have a lot a comment the code maybe?
One thing I have in my todo list is to replace the `dateDisplay` documentation page with a `format` dedicated to Date. What do you think?

packages/svelte-ux/src/routes/customization/+page.md
```diff
@@ -80,6 +80,22 @@ settings({
         fractionDigits: 4,
       },
     },
+
+    dates: {
+      // This is the default, but you can override it here for your app
+      locales: 'en',
+      weekStartsOn: 0, // Sunday
```
techniq, line 87: Should we use `DayOfWeek.SUN` here?

packages/svelte-ux/src/routes/+layout.svelte
```diff
@@ -26,13 +26,26 @@
   $: title = data.pr_id ? `🚧 (pr:${data.pr_id}) - ${baseTitle}` : baseTitle;
 
   settings({
+    // Usefull to test different locales with the docs
```
techniq, line 29: See comment about adding global settings dialog (in future).  Will help a lot with this

techniq (commented):
Some quick comments before making the kids lunches.  Looking REALLY good!

techniq:
Few changes I noticed (mostly just tracking, not saying they need to be changed yet).

## DateRangeField

The range now shows numeric instead of month abbreviations (`12/29/2023` instead of `Dec 29, 2023`).  It also shows leading `0`s for months/days (`01/01/2024` instead of `1/1/2024`).  Neither are major, and I'm sure we can change/tweak the default for some of these cases.  I also need to think if it's better or worse than before :).

Before | After
--- | ---
image | image

## Month

Days of the week now show better abbreviations (`Mon`/`Tue` vs `M`/`T`).  Thank you :)

Before | After
--- | ---
image | image


One thing I would love to do after this (and design token is merged) and allow you to change the locale/`settings()` in the docs and have everything update.  Similar to how you can change the theme via the AppBar.  I'm debating on whether this should be a modal/dialog/drawer or separate page, and whether it should be an extension of the current theme selector or separate.  Anyways, I'd rather we merge both of these PRs before we do that work since they would both be impacted.

packages/svelte-ux/src/lib/utils/date.ts
jycouet: Most of my code is camelCase, but the more I see snake_case, the more I like it. I know that for consistency, it could be better to go back to camelCase.

I feet it a bit strange with `dayOfMonth2Digit` (because au the 2)
I was also thinking about: `dayOfMonth_2Digit_dd` to give as much info as possible with just the name. like: `meaningFullName_intlInfo_fnsFormatFYI`

Let me know what's your final, though, I'll adapt to it 👍

techniq: I don’t mind snake case (full disclosure, I wrote a lot of Python about 10 years ago). I also believe the Svelte project uses it (Rich’s preference if I recall). With that said, I’d rather be consistent and maybe one day switch. I do prefer camel case for props (and aligns with DOM apis more). I’m fine with using them for meaningful dividers in this case.

So in short, prefer camel case but can use underscores with enough reason.   Sound good to you?

packages/svelte-ux/src/lib/components/settings.ts
```diff
@@ -49,3 +57,147 @@ export function getFormatNumberOptions(style?: FormatNumberStyle) {
...
+    custom: CustomIntlDateTimeFormatOptions;
+    presets: {
+      days: Record<DateFormatVariant, CustomIntlDateTimeFormatOptions>;
+      weeks: Record<DateFormatVariant, CustomIntlDateTimeFormatOptions>;
+      months: Record<DateFormatVariant, CustomIntlDateTimeFormatOptions>;
+      monthsYears: Record<DateFormatVariant, CustomIntlDateTimeFormatOptions>;
+      years: Record<DateFormatVariant, CustomIntlDateTimeFormatOptions>;
+    };
+    ordinalSuffixes: Record<string, OrdinalSuffixes>;
+    dico: {
```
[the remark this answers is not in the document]
jycouet: >  assume dico is short for dictionary? I haven't seen this abbreviation if so. Could we call it dictionary or translation?

Oups, yes I'm use to use `dico`, but it's maybe not so common ^^.
I feel that `translation` is like: 
```ts
translation: {
  en: {
    "day": "Day"
  }
}
```
`dictionary` is more:
```ts
dictionary: {
  "day": "Day"
}
```
Saying that, it's maybe better to go to `translation` (more future proof? fallbacks, ...)

---

True, that I will need also `global` things (less urgent).

Is it better to start directly under another place beside `formats`? or later? or?
(I would like to not have a too big PR, but would like also to not break by moving things around 2 weeks after introducing them ^^)

Let me know

packages/svelte-ux/src/routes/customization/+page.md
jycouet: Yes, good point 👍
I'll update it

packages/svelte-ux/src/lib/components/DateRangeField.svelte
```diff
@@ -104,7 +103,7 @@
     on:click={() => (open = true)}
     {id}
   >
-    <DateRangeDisplay {value} {format} />
+    <DateRangeDisplay {value} />
   </button>
```
jycouet, line 107: I put this here, to track it. I'll look later (there are maybe other places). It's cool to check it out with prod doc & PR doc 🎉

> The range now shows numeric instead of month abbreviations (`12/29/2023` instead of `Dec 29, 2023`).  It also shows leading `0`s for months/days (`01/01/2024` instead of `1/1/2024`).  Neither are major, and I'm sure we can change/tweak the default for some of these cases.  I also need to think if it's better or worse than before :).

Before | After
--- | ---
image | image

packages/svelte-ux/src/lib/components/settings.ts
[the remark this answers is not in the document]
techniq: TBH, I'm not an expert here.  I'm hoping with Svelte UX / LayerChart to push as much of this into `Intl.*` browser APIs or similar to not get in the way, but also not have to maintain it.

With that said, I know these kind of strings still need to be handled.  At first I thought using `settings()` (i.e. `context`) might be an issue with SSR (will we always render with English, but when it hydrates in the browser and switches to say French, will there be a flash, or can we still set French on the backend by reading `Accept-Language` or a user preference from the DB).  Based on the SvelteKit [docs](https://kit.svelte.dev/docs/state-management#using-stores-with-context), I don't think context will be an issue here, so maybe I'm overthinking it.

I assume to do i18n properly you would want to use something like [typesafe-i18n](https://github.com/ivanhofer/typesafe-i18n) or [Paraglide JS](https://inlang.com/m/gerre34r/library-inlang-paraglideJs) with some preprocessing, but I don't think that is something we would want as a dependency within this project...

All that's to say, I think this works for me, but I could see some additional iteration in the future, especially with people with more direct needs / understanding (we don't need to boil the ocean now).  My hope is to keep the number of translatable strings within Svelte UX and LayerChart under 10 :).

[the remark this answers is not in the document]
jycouet: > SSR
I didn't look so much into this (I will at some point). But maybe not in this PR. I think that it should be ok with the current setup, would b

packages/svelte-ux/src/lib/components/DateRangeField.svelte
jycouet: Hooo, yes, now I remember ^^.
Because in the `days.default`, I've set by default default `baseParsing` that is: 'MM/dd/yyyy'.
I can remove this, behavior, be we need to test that parsing is still working good.

I'll revert this, we will see if parsing will be an issue.

packages/svelte-ux/src/lib/components/settings.ts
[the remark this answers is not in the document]
techniq: Let's use top-level `translations` for now to allow to grow (and now have to come up with a "namespace" for some things like button labels.

packages/svelte-ux/src/lib/utils/date.ts
techniq: Looks good.  Thanks!

techniq:
Once this is ready, I would like to publish as `@next` to update `LayerChart` (like the design token PR).  I'll probably need to stack/merge both onto a `next` branch, since pushing to `main` will currently publish to the main docs as well.  I'm hoping to get the design PR to a near-mergable state by the weekend, although I return back to work tomorrow and have evening kid activities (basketball games, etc) so my availability will be spotty.

I can't say this enough, this is looking great, and a huge improvement!  Will need to kick it around a little to wrap my head around it fully, but really looking forward to it.

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
```diff
@@ -83,13 +102,277 @@
...
+<h1>Dates</h1>
+
+<h2>No format</h2>
+
+<Preview>
+  {format(myDate)}
+  <!-- TODO JYC?
+    before: 30/03/1982 00:00:00
+    after: Tue Mar 30 1982 00:00:00 GMT+0200 (heure d’été d’Europe centrale)
+  -->
```
jycouet, line 114: I'm not sure what to do here. It's maybe the "only" difference with `dateDisplay`!
I would remove this example. 
Ok with that?

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
```diff
@@ -83,13 +102,277 @@
...
+    <h3>short</h3>
+    <Preview>
+      {format(myDate, PeriodType.Day, {
+        variant: 'short',
+      })}
+    </Preview>
+  </div>
+  <div>
+    <h3>default</h3>
+    <!-- TODO JYC QUESTION: before: 3/30/1982, suggestion as default: 03/30/1982 -->
```
jycouet, line 159: It's back to what exist today in prod, I wonder if it's not better with leading 0 (even in `en`)?
I let you tell me ^^

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
```diff
@@ -83,13 +102,277 @@
...
+<div class="grid grid-cols-3 gap-4">
+  <div>
+    <h3>short</h3>
+    <!-- TODO JYC QUESTION: before: Mar, suggestion as default: 3 -->
+    <Preview>
+      {format(myDate, PeriodType.Month, {
+        variant: 'short',
+      })}
+    </Preview>
+  </div>
```
jycouet, line 255: What do you think about this change?
Just a question to have a bit better default default. Let me know

packages/svelte-ux/src/lib/components/settings.ts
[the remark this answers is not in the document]
jycouet: Ok, I did something. It will be easy to add https://inlang.com/ in userland.
I added `Ok` and `Cancel`, just to demo. But I don't expect to expend here in this PR.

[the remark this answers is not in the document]
techniq: Looks good, thanks!

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
techniq: We can remove the example as the output is the same as `date.toString()`.

image

It was nice to have a quick way to show a consolidated date/time, but not a big deal.  ATM we don't handle any time formatting, although I used `dateDisplay` for one example on LayerChart

image

I also use date-fns's `format()` for a few other time formats.

image

I figure it would be nice to have an intl. way to handle time as well.  Should we add `PeriodType.DayTime` (or `DateTime`) and `PeriodType.TimeOnly` (or just `Time`) with short and long variants?

- `PeriodType.DayTime`
  - short: 'M/d/yyyy hh:mm a'
  - long: 'M/d/yyyy hh:mm:ss a'
- `PeriodType.TimeOnly`
  - short: 'hh:mm a'
  - long: 'hh:mm:ss a'

Although this brings up a few time specific questions?
- Should we show as 24 hour or 12 hour am/pm.  I figure this is locale specific in most cases (US is am/pm typically)
- Do we only show seconds for `long`?  what about milliseconds (`SSS`).  Probably just handle via `custom`, but would be good to support it if possible (when I look at the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), I don't see any mention of milliseconds, so maybe not possible... hmm.  If not, using date-fns for these use cases is fine.

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
techniq: I'd prefer non-leading 0s by default.  It's also the default for `new Intl.DateTimeFormat("en-US").format(...)`

image

techniq: I won't say there isn't a case for leading 0s... lining up dates in a table column is so much nicer with leading 0s, but also I think it's a little easier to read without, and when you are space constrained, it can be nice to not have 0s as well.  At least it's configurable :)

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
techniq: I think it should still be "Mar" for short and "March" for long.  If only showing the month by itself, only showing a number I think can be confusing, or am I overlooking a use case?

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
jycouet: Great points. Wanted to push this out for later... But in the end, it's pretty easy to handle all these cases!

With `custom` you can already do all that, but yes, let's add `DayTime` & `TimeOnly`.
`intl` supports also `24 & am/pm` if I recall well. I will check.

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
jycouet: `new Intl.DateTimeFormat("fr").format(new Date())` has leading 0 ;)

It's all configurable, so it's OK. but find it "strange" to have a default parsing different than a default display.
I'll remove the comment and resolve here 👍 (until it bugs me more ^^)

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
jycouet: We have 3 levels now, today it's: 
```ts
{
  'short': "Mar"
  'default': "Mar"
  'long': "March"
}
```

That's why I was wondering to have another default default for `short` as `Mar` is already covered in `default`.

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
techniq: It does via `hour12` (I saw when I was looking for millisecond support, which I didn't see)

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
techniq: Awe, good point... hmm.  It is odd to have 2 result in the same.  I'm ok with `default` returning `3` in this case then.  Thanks

packages/svelte-ux/src/routes/docs/utils/format/+page.svelte
jycouet: All Done 🎉

(I resolve this one, let's open new one if needed)

techniq:
@jycouet I'll give it a final review this evening (I'm booked up this afternoon with meetings and need to get some head down work done).

We'll also need to merge this into the `next` branch instead of `main`.  I actually [already](https://github.com/techniq/svelte-ux/tree/next) created it a few days ago (but didn't end up using it (yet) for the design token PR, but I'll restack on top of it once you merge to it).  It's a few commits behind at the moment if you wouldn't care to pull/sync it up.

jycouet:
> @jycouet I'll give it a final review this evening (I'm booked up this afternoon with meetings and need to get some head down work done).
> 
> We'll also need to merge this into the `next` branch instead of `main`. I actually [already](https://github.com/techniq/svelte-ux/tree/next) created it a few days ago (but didn't end up using it (yet) for the design token PR, but I'll restack on top of it once you merge to it). It's a few commits behind at the moment if you wouldn't care to pull/sync it up.

I let you have a look at `git` fun :p

techniq:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/techniq/svelte-ux/137?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

Current dependencies on/for this PR:
* `main`
  * **PR #167** <a href="https://app.graphite.dev/github/pr/techniq/svelte-ux/167?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #137** <a href="https://app.graphite.dev/github/pr/techniq/svelte-ux/137?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/techniq/svelte-ux/137?utm_source=stack-comment).

PR #141, opened 2023-11-24
fix: keyboard shortcut conflict for opening QuickSearch

shyakadavis:
I use Firefox, and `cmd+k` shortcut, by default, is used to focus the search bar - [ref](https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly). (Or just test out the site in Firefox)

As such, they always conflict with Svelte-UX's.

This is a fix for that.

Also, I added support for `ctrl+k` on Windows. I don't have a way to test this for Windows, but I'm confident it should work.

Although, there is one thing I couldn't figure out. When I hit `cmd+k`, it works fine, but when I hit `cmd+k` again, it doesn't close the dialog, but this time jumps to the browser's default, and in my case, focuses the search bar. I thought adding `open != open` would fix this, but it didn't. Does the `keyDown` function only apply once and can't be called again? 🤔 

Let me know what you think, @techniq.

### DIFF
--- a/packages/svelte-ux/src/lib/components/QuickSearch.svelte
+++ b/packages/svelte-ux/src/lib/components/QuickSearch.svelte
@@ -29,8 +29,9 @@
   */
 
   function onKeyDown(e: KeyboardEvent) {
-    if (e.key === 'k' && e.metaKey) {
-      open = true;
+    if (e.key === 'k' && (e.metaKey || e.ctrlKey)) {
+      e.preventDefault();
+      open = !open;
     }
   }
 </script>


techniq (approved):

techniq:
Thanks @shyakadavis.  I looked briefly into the "Command+K" not closing the dialog.  I thought it was due to the Dialog's [stopPropagation](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Dialog.svelte#L98) causing it not to bubble, but commenting it out didn't fix it.

This `stopPropagation()` is to keep clicking escape with multiple modals (menu, dialog, drawer, etc) from reach the next level down.  It could probably me moved within the `if (e.key === 'Escape') {` block but since it didn't fix this issue, I just left it for now.

I'm going to merge this as is so Firefox users can start to benefit from it, and i figure closing via this (verse `escape` is probably a low priority.

Thanks for the PR!

PR #151, opened 2023-12-07
Add slot option for AppBar

cyates81:
If you add a slot with the name title it will override the prop but the prop can still update the window title.

### DIFF
--- a/.changeset/sweet-pens-change.md
+++ b/.changeset/sweet-pens-change.md
@@ -2,4 +2,4 @@
 'svelte-ux': patch
 ---
 
-Add slot option for AppBar title.
+[AppBar] Add slot option for title.


techniq:
@cyates81 I've been meaning to do this, thanks!  Could you run `pnpm changeset` to produce a [changeset](https://github.com/changesets/changesets) so this shows up in the changelog?  I would use `patch` (as pre-1.0 I treat non-breaking changes including features as patch, and breaking changes as `minor`.  Post `1.0` we'll follow standard semver practices).  Thanks again!

techniq (approved):

techniq:
Thanks @cyates81.  Published as `0.57.1`

cyates81:
You're welcome! Thank you for your hard work on the library. Still in the initial testing phase but we are hoping to use it for our project. If we do and come across enhancements we would need that others might benefit from I'll do PRs.

Issue #132, opened 2023-11-14
[SelectField] Standardize with MenuField / general improvements


techniq:
Extracted from PR #130

- [x] Use same [Options](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/MenuField.svelte#L15) type instead of [any](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/SelectField.svelte#L25)
  - Will use `label` instead of `name` by default (can still be overwritten by `optionText()`, maybe we call this `optionLabel()`, or maybe we require the data to be processed ahead of time and remove this and `optionValue()`
- [ ] Pass `icon` prop to [MenuItem](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/SelectField.svelte#L447-L461) like [MenuField](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/MenuField.svelte#L130)
- [x] Should we update the `<SelectField icon={...} />` based on the selected `<MenuItem icon={...} />`?
  - Add a prop to opt-in to this functionality (`activeOptionIcon` or something)
  - Require the user to do this (like the SelectField [example](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/routes/docs/components/SelectField/+page.svelte#L175) currently does.
- [x] Fix clicking on Chevon icon/button to close (re-toggles open) (#93)

techniq:
The first task is also in the [Road to v1](#59) issue

techniq:
Task 1 and 3 resolved via https://github.com/techniq/svelte-ux/pull/127.

techniq:
@willnationsdev Tagging you here if you're interested in knocking out these, but don't feel like you have to.

willnationsdev:
Got a PR, but even with the icons added back in, the blue coloration they had before on the docs page is now gone. I'd bet there're styles somewhere in the docs that need to be adjusted. Cause you wouldn't wanna make a change like that to the component itself or to the theme that applies to everyone's usage of the SelectField's MenuItems - you just want the docs page. But the docs page didn't appear to have Tailwind classes is was merging in via attributes, so idk where those styles came from.

Issue #93, opened 2023-09-29
[SelectField] Clicking chevron arrow re-opens menu


techniq:
Clicking chevron arrow while open does not close, but re-opens quickly (on mouse up)

https://github.com/techniq/svelte-ux/assets/177476/4c153eb4-82e5-44fb-8b10-358d60c8f5b4

techniq:
This was fixed a while back

Issue #91, opened 2023-09-21
[MenuField] Re-opening menu while transition close causes `classes` to be undefined (instead of default)


techniq:
If you quickly click to re-open the menu while it's transitioning closed,`classes` within `MenuItem` will be set to `undefined` instead of default object/styles.

https://github.com/techniq/svelte-ux/assets/177476/a36c772d-0717-4494-8909-c4674aa288a0

- [MenuField](https://github.com/techniq/svelte-ux/blob/master/src/lib/components/MenuField.svelte#L133)
- [MenuItem](https://github.com/techniq/svelte-ux/blob/master/src/lib/components/MenuItem.svelte#L16-L20)

Within `MenuField`, If you don't pass `classes` to `MenuItem` or set it explicitly to `undefined` (`<MenuItem classes={undefined} />`, the issue does not present.

techniq:
This also occurs if you click on the chevron toggle button to close a menu already opened.  See https://github.com/techniq/svelte-ux/pull/155#issuecomment-1864652385 for some context.

Would be good to test with Svelte 5 to see if this is still an issue.  For the chevon button issue, ultimately it would be nice if the menu was closed and not quickly re-opened. See issue #93

PR #155, opened 2023-12-19
Fix SelectField MenuOption icons and non-collapsing toggle button.

willnationsdev:
Related to #132.

### DIFF
--- a/.changeset/bright-donuts-appear.md
+++ b/.changeset/bright-donuts-appear.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+Fix SelectField MenuOption icons no longer being passed to MenuItems. Fix SelectField `classes: { field: ... }` not being incorporated into the inner `TextField` properly. Same with `classes: { option: ... }` not being incorporated into each inner `MenuItem` option. Update SelectField docs page with colored example that differentiates icon & input color, both in the field and the menu items.
--- a/packages/svelte-ux/src/lib/components/SelectField.svelte
+++ b/packages/svelte-ux/src/lib/components/SelectField.svelte
@@ -67,14 +67,41 @@
     root?: string;
     field?: string | ComponentProps<TextField>['classes'];
     options?: string;
-    option?: string;
+    option?: string | ComponentProps<MenuItem>['classes'];
     selected?: string;
     group?: string;
     empty?: string;
   } = {};
 
   let fieldClasses: ComponentProps<TextField>['classes'];
-  $: fieldClasses = typeof classes.field === 'string' ? { root: classes.field } : classes.field;
+  $: {
+    let v: {root?: any} = {};
+    if (typeof settingsClasses?.field === 'string') {
+      v = { root: settingsClasses.field };
+    } else {
+      v = settingsClasses?.field ?? {};
+    }
+    if (typeof classes?.field === 'string') {
+      fieldClasses = { ...v, root: cls(v.root, classes.field)};
+    } else {
+      fieldClasses = { ...v, ...classes?.field ?? {} };
+    }
+  }
+
+  let optionClasses: ComponentProps<MenuItem>['classes'];
+  $: {
+    let v: {root?: any} = {};
+    if (typeof settingsClasses?.option === 'string') {
+      v = { root: settingsClasses.option };
+    } else {
+      v = settingsClasses?.option ?? {};
+    }
+    if (typeof classes?.option === 'string') {
+      optionClasses = { ...v, root: cls(v.root, classes.option)};
+    } else {
+      optionClasses = { ...v, ...classes?.option ?? {} };
+    }
+  }
 
   // Menu props
   export let placement: Placement = 'bottom-start';
@@ -425,11 +452,13 @@
     on:keypress={onKeyPress}
     actions={fieldActions}
     classes={{
+      ...fieldClasses,
       container: inlineOptions
-        ? 'border-none shadow-none hover:shadow-none group-focus-within:shadow-none'
-        : undefined,
+        ? cls('border-none shadow-none hover:shadow-none group-focus-within:shadow-none', fieldClasses?.container)
+        : fieldClasses?.container,
+      
     }}
-    class={cls('h-full', settingsClasses.field, fieldClasses)}
+    class={cls('h-full')}
     role="combobox"
     aria-expanded={open ? 'true' : 'false'}
     aria-autocomplete={!inlineOptions ? 'list' : undefined}
@@ -534,6 +563,7 @@
           <svelte:fragment slot="option" let:option let:index>
             <slot name="option" {option} {index} {selected} {value} {highlightIndex}>
               <MenuItem
+              	classes={optionClasses}
                 class={cls(
                   index === highlightIndex && '[:not(.group:hover)>&]:bg-surface-content/5',
                   option === selected && (classes.selected || 'font-semibold'),
@@ -549,6 +579,7 @@
                 role="option"
                 aria-selected={option === selected ? 'true' : 'false'}
                 aria-disabled={option?.disabled ? 'true' : 'false'}
+                icon={option.icon}
               >
                 {optionText(option)}
               </MenuItem>
@@ -594,6 +625,7 @@
         <svelte:fragment slot="option" let:option let:index>
           <slot name="option" {option} {index} {selected} {value} {highlightIndex}>
             <MenuItem
+              classes={optionClasses}
               class={cls(
                 index === highlightIndex && '[:not(.group:hover)>&]:bg-surface-content/5',
                 option === selected && (classes.selected || 'font-semibold'),
@@ -609,6 +641,7 @@
               role="option"
               aria-selected={option === selected ? 'true' : 'false'}
               aria-disabled={option?.disabled ? 'true' : 'false'}
+              icon={option.icon}
             >
               {optionText(option)}
             </MenuItem>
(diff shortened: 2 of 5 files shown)


techniq:
Interesting.  Not sure why Cloudflare isn't producing a PR preview deployment for this, but they show up for other PRs.

image

techniq:
Well crap...

image

techniq:
@willnationsdev I re-setup Vercel deployments (and switch to the SvelteKit auto adapter).  Can you sync this branch with  `main` and push to hopefully trigger a PR preview (not a major issue for this PR, and I would think would be fixed for new PRs opened).  Going to do the same for LayerChart later

techniq:
This PR actually wasn't a regression but an improvement :).  With that said, could you update the changeset?

Also, you use update this example (with a different name) to leverage options with `icon` set.  I think we won't need to ever set the `option` slot anymore...

image

We might need to adjust passing `classes` to `MenuItem` to style the `icon`, but I'm not overly concerned with that tbh (you can still override the slot for these more "complex" use cases).  I just thought it would be nice to have an example using `option.icon` set to show the simplicity of setting these icons.

Thanks!

willnationsdev:
I updated the example in the docs so that it shows how to individually control both the text & icon color within both the field & the menu items. I also noticed, in the process, that the way I was passing the `fieldClasses` into the `TextField` was incorrect (I now understand that `cls(...)` only knows how to build up a single class string, not how to merge objects that track multiple class strings - that's what `classes` is for).

willnationsdev:
Also, I'm not really sure, but is the `placements` example at the bottom of the SelectField docs page working correctly? The placement of `top-start` isn't really cluing me in to whether or not it's doing something cause it seems to more or less behave like the other simple examples. Wondering if merging from the SelectList caused any issues there.

techniq:
> Also, I'm not really sure, but is the `placements` example at the bottom of the SelectField docs page working correctly? The placement of `top-start` isn't really cluing me in to whether or not it's doing something cause it seems to more or less behave like the other simple examples. Wondering if merging from the SelectList caused any issues there.

Good catch, but it wasn't a regression with your changes :).  Here is a [preview](https://b92b00a0.svelte-ux.pages.dev/docs/components/SelectField/) before the merge for reference.

I think the issue is I changed [autoPlacement](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/SelectField.svelte#L73) to true by default, which picks the side of the viewport with the most space.  `placement` is mostly a suggestion and based on which `floating-ui` modifiers are enabled determine the final location.

Regarding the button toggle, I see you currently [ignore](https://github.com/techniq/svelte-ux/pull/155/files#diff-068539603e2e3194353cb2317456128688f67a40892e7d430d6b68e7f7d1a724R240) the click if press it, which I think feels wrong (and should toggle close). 

https://github.com/techniq/svelte-ux/assets/177476/ed0b75d5-9be4-4ba0-beec-52997b295ab2

It also reveals an odd issue I've seen that I think is a Svelte bug (hoping Svelte 5 just fixes it, but haven't had time to test).  I think the issue is initial prop values are not set when something transitions in/out quickly (at least that's ben the best I could figure out.  Basically `MenuItem`'s default [classes](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/MenuItem.svelte#L17-L21) are being lost (not the font size change in the video).

What if we dont worry about this toggle issue ATM and ship the other improvements, and we can circle back later.  If you're OK with that, would you mind backing our those changes (including the new `bind:this` props on Button).  If you still want to try to get this resolved, I'm fine with that too, but I didn't want to hold up getting you a release out.

willnationsdev:
> Regarding the button toggle, I see you currently [ignore](https://github.com/techniq/svelte-ux/pull/155/files#diff-068539603e2e3194353cb2317456128688f67a40892e7d430d6b68e7f7d1a724R240) the click if press it, which I think feels wrong (and should toggle close).

The issue is that the toggle button triggers and I flip the `open` state accordingly, but then the `onClick` and `onFocus` handlers for the `TextField` execute *afterward* and end up forcing it back to `open = true`. If it were just a click handler alone, I could use `stopPropagation()` to prevent the event from making it to the `onClick` event, but I couldn't figure out how to make the toggle button prevent *other* kinds of events from being triggered on the ancestor; so instead, I borrowed from the strategy you were already employing in [other parts of the code](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/SelectField.svelte#L217) and directly filtered out the toggle button's elements from being a possible `relatedTarget` of the `MouseEvent` or `FocusEvent`.

> What if we don't worry about this toggle issue ATM and ship the other improvements, and we can circle back later. If you're OK with that, would you mind backing our those changes (including the new bind:this props on Button). If you still want to try to get this resolved, I'm fine with that too, but I didn't want to hold up getting you a release out.

That's fine too. I've pushed an update that removes those changes, but I also have a local backup if you end up wanting it restored. Although, if you have another solution in mind that I didn't think of, then I'm all ears since I'm still pretty intermediate-level when it comes to frontend web development.

techniq:
I appreciate you taking a stab at it, and that might be an approach we can leverage.  I recall taking a few attempts at it, hoping to leverage `stopPropagation` / `stopImmediatePropagation`, but the focus changing in/out made that difficult IIRC.  

I might be able to use some kind of setTimeout() "cooloff" to handle the toggling, similar to how [Tooltip](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Tooltip.svelte#L40-L55) changes the delay based on any if Tooltip was shown in the last 500ms (not exactly the same thing or how we would handle it, but a possible approach).

I think for the time being I'd rather leave it as is (a little simpler) until I have a little more bandwidth to really think about it, if you're OK with that :).

Also, I just realized I already have most of this tracked in #91 and #93, so if you're happy, I'm good with merging this PR and closing #132, and tracking those issues separately.

willnationsdev:
Yeah, that sounds like a good plan to me. Thank you so much for being responsive/active/advisory with regards to all these changes. :-D

willnationsdev:
@techniq I actually came up with a much smoother & simpler solution that solves it in a similar vein as before, but without being quite so picky about the elements involved or having to extract elements from other components. [Check it out](https://github.com/techniq/svelte-ux/commit/d603ea10e7aa4974bb063b6f3fe1492182363054), and let me know if you'd like me to add it to this PR.

techniq:
@willnationsdev Welcome, I try my best :)

I like that approach much better, although it looks like it still would cancel the closing of the menu if clicked on the toggle icon/button, and I feel that icon should always toggle open or close based on the current state.

Btw, Popover has a similar "clickOutside" [check](https://github.com/techniq/svelte-ux/blob/62c02d4029aebeeda43c38606eaa752ea5298d52/packages/svelte-ux/src/lib/actions/popover.ts#L82-L91) based on 2 elements (anchor/trigger element, and the popover itself).  I first thought I had a `clickOutside` action, but remembered it typically needed more than a single element to check, which is the typical use case for the action.  We could still create an action and pass a secondary element for reference, if we decided to implement it at some point.   I know `clickOutside` was discussed being added directly to Svelte in an old [RFC](https://github.com/sveltejs/rfcs/pull/24), but it kind of hit a block. 
 Swyx did create a repo of some, including [this](https://github.com/swyxio/svelte-actions).

With that said, do you think clicking the toggle icon/button should still open/close based on the current state?

willnationsdev:
> I like that approach much better, although it looks like it still would cancel the closing of the menu if clicked on the toggle icon/button, and I feel that icon should always toggle open or close based on the current state.
> 
> With that said, do you think clicking the toggle icon/button should still open/close based on the current state?

It still opens/closes based on the current state. The problem was that when clicking on the toggle button to close, the `onBlur` event would trigger *before* the click handler executed. It would set `open` to false preemptively *just beforehand*, so that when the toggle click event handler runs, it "flips" it *back* to being open.

Although, *now* I'm finding that the blur event is never properly hiding the menu if you tab away from the SelectField, so some adjustment is still necessary, even if it's moving in a better direction. Something about detecting which area was clicked on & then leveraging that.

techniq:
@willnationsdev I forget where we left this one exactly.  I believe you were investigating the chevron clicking issue and improving that interaction.  Once #167 is merged (hopefully in the next 2 weeks), could you rebase this PR and we'll what remains?  Thanks!

willnationsdev:
Yeah, sounds good. I'll do that.

techniq:
@willnationsdev `next` has now been merged so you can rebase when you're ready.  Thanks!

willnationsdev:
> @willnationsdev `next` has now been merged so you can rebase when you're ready.  Thanks!

Okay, will do. Might be later tonight, but I'll do it ASAP.

techniq:
No rush :)

techniq:
Thanks @willnationsdev.  I ran `pnpm format` locally and pushed to your branch to fix the build complaining.  Hopefully the `.editorconfig` I pushed last night fixes your local Windows environment.  Let me know if not and I'll take another look.

I did a quick comparison and noticed the menu items are no longer receiving their `text-sm` class...

Before | After
--- | ---
image | image

Related note: I need to take a good hard look at how we apply class/classes across all components with have a simple/repeatable process (without introducing any regressions).  I started tracking this in #245 for reference.

techniq:
@willnationsdev It's because of how `classes` is [set](https://github.com/willnationsdev/svelte-ux/blob/pr/willnationsdev/155/packages/svelte-ux/src/lib/components/MenuItem.svelte#L20-L24) as the default value of the prop of none is set.  Hmm... might have to think about #245 sooner than I hoped :)

techniq:
@willnationsdev Going to close this for now, but feel free to re-open

PR #146, opened 2023-11-30
Design tokens / Theme support (with dark mode)

techniq:
Nothing to see here...
backing-up-homer-simpson-gif

Resolves #66

PR #158, opened 2023-12-21
Add renovate

jycouet:
I would suggest to add this default. (you can look at details here: https://github.com/jycouet/jycouet/blob/main/shared-config/renovate.json)

It's
- [x] Extending their `config:base` 
- [x] Reducing noice (only minor) 
- [x] add a few groups for svelte

## The best would be to:
1/ Merge this **FIRST** _(to pick up directly the good default)_
2/ Install https://github.com/apps/renovate for this repo
3/ Enjoy 🎉

## What you will get?
- An issue that will auto update for you to get an overview of what is going on. (usually I pin it)
- Some PRs coming from time to time

## What if?
- It's too much? to noisy? => We can tweak configs 👍
- It's really not for you? => We can remove the file `renovate.json`, remove the app, and renovate is gone!

I think that it's worth trying ✅

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,8 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["github>jycouet/jycouet//shared-config/renovate/default.json5"],
+
+  // Avoid noise in the middle of the day
+  "timezone": "Etc/Greenwich",
+  "schedule": ["after 4am and before 5am"]
+}
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,8 +1,46 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["github>jycouet/jycouet//shared-config/renovate/default.json5"],
+
+  // Extend the base config
+  "extends": ["config:base", ":semanticCommitTypeAll(chore)"],
+
+  // Some tuning that works well for my repos
+  "automerge": false,
+  "rangeStrategy": "replace",
+  "baseBranches": ["main"],
+
+  // For a nice dashboard
+  "dependencyDashboardTitle": "📦 Dependency Dashboard",
+  "labels": ["dependencies"],
 
   // Avoid noise in the middle of the day
-  "timezone": "Etc/Greenwich",
-  "schedule": ["after 4am and before 5am"]
+  "timezone": "EST",
+  "schedule": ["after 1am and before 5am"],
+
+  "packageRules": [
+    // Let's update each patch for these libs
+    {
+      "matchUpdateTypes": ["patch"],
+      "enabled": false,
+      "excludePackagePrefixes": ["layerchart", "svelte-ux"]
+    },
+    // devDependencies will be prioritized last
+    {
+      "matchDepTypes": ["devDependencies"],
+      "prPriority": -1
+    },
+    // Groupings
+    {
+      "groupName": "Svelte",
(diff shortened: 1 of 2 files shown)


techniq:
I think this is a great idea!

Let me get #146 closer to release (at least as `next`) before we enable this, just so I don't lose my focus.  I know I need to bump some deps already (SvelteKit 2, etc), but wanted to get that PR on `main` before I did so.

techniq:
I would also like to add the full [config](https://github.com/jycouet/jycouet/blob/main/shared-config/renovate.json) in this repo instead of as an `extends` reference so I can modify it directly :)

jycouet:
> I think this is a great idea!
> 
> Let me get #146 closer to release (at least as `next`) before we enable this, just so I don't lose my focus. I know I need to bump some deps already (SvelteKit 2, etc), but wanted to get that PR on `main` before I did so.

As you want, but you don't need to be up-to-date on everything to start & see. _(BTW, if you update main with something else, `renovate PRs` will update themself 😉)_. You can also totally ignore PRs or even close them to say to renovate: "please don't update this one", it will appear in a dedicated section in the `📦 Dependency Dashboard`.

> I would also like to add the full [config](https://github.com/jycouet/jycouet/blob/main/shared-config/renovate.json) in this repo instead of as an extends reference so I can modify it directly :)

I totally get it, yes 👍. _At the same time, you will not copy locally their `"extends": ["config:base"]` 😜_
The idea was to share common practices and evolve config together on grouping. And locally, you can also add some dedicated things on top.
FYI: On my repos, I would be also hesitant to extend not the base, but someone else! 
Let me know your final decision, I will update accordingly, _(no issue for both options 🎉👍)_

techniq:
Sounds good.  I do think I'd rather have a local copy of the config, as I might also use it as a basis for my other repos (LayerChart being the most obvious).  While I know I could extend this one that extends yours, I'd like to stay as close to the source as possible.

I looked briefly as renovate and dependabot and they both look similar, but it does appear renovate is a little more flexible. I see dependabot supports grouping as well (but all manual ATM).  Part of me is wanting to use dependabot because there is a much lower change of it disappearing one day, but the extra flexibility in renovate looks like it's worth it.  We can always switch later if needed (just have to figure out how to do the same with the other tool, and likely will miss a feature or way a workflow the tool provides).

With that said, lets use renovate.  If you wouldn't care to change the config to use `config:base` directly , and feel free to use any of the groupings you have in yours if you want (we won't need things like Svelte UX & LayerChart, Houdini, and Envelop) and change the timezone to EST :), I'll activate it and see how it goes.  Thanks!

jycouet:
Cool feedback, we are not in hurry, I'll do the change following what you said 👍

techniq:
Thanks for improved all the QOL improvements lately :)

jycouet:
Finally back, here is the update to local things, probably a good base?

techniq (approved):
LGTM.  Thanks @jycouet

techniq:
Merged and setup.  A lot coming on the next schedule (this evening).

image

I do know we'll need to hold off on the vitest update (`0.34`) as I had to roll back a few months ago due to this [issue](https://github.com/vitest-dev/vitest/issues/3886).  It's still open but looks like directly installing [@sveltejs/vite-plugin-svelte](https://github.com/vitest-dev/vitest/issues/3886#issuecomment-1745058584) should fix the issue.

jycouet:
It's ok, not everything will be created, it's just the list 👍
For the one you know like `vitest`, you can just close the PR and it will show in a different section in the issue.

PR #167, opened 2024-01-03
Next release

techniq:
## TODO

- [x] Map Skeleton themes to Svelte UX themes, similar to Daisy UI themes.
  - [ ] Determine how this will work when using Skeleton with LayerChart (and not Svelte UX).  "Should work" 🙂
- [x] Determine best way to publish theme mappings
  - [x] ~~Create `@svelte-ux/themes` CLI that generate a `theme.json` file that can then be read in `tailwind.config.js`(depends on `daisyui`/`skeleton` and not the main project)~~
  - [x] Online theme generator that produces JSON (#192)
- [x] Theme browser / editor
  - [x] Browse Daisy, Skeleton, etc themes and pick one to use for `light` and `dark` themes
  - [x] Ability to edit (or create a new one from scratch)
  - Related:
    - https://www.skeleton.dev/docs/generator
    - https://daisyui.com/theme-generator/
- [x] Safari error: `Unhandled Promise Rejection: ReferenceError: Cannot access uninitialized variable.`
  - Fix: https://github.com/techniq/svelte-ux/pull/214
- [x] Fix Bi-Weekly `DateRange` regression ([discussion](https://discord.com/channels/920755200552226868/1192979604445614150/1194742214518132776))
  - https://github.com/techniq/svelte-ux/pull/217
- [x] Period type only changes (Month => Day) are not updating the selection
- [x] Finish LayerChart design tokens PR (review remaining examples, etc)
- [ ] Dogfood within personal projects
  - [x] Strava Analysis
  - [x] Github Analysis
  - [ ] Personal site
  - [ ] Work project
- [ ] Update `create-svelte-ux` templates
- [x] Extract theme settings (localStorage, <html data-theme="..." class="dark" /> from ThemeButton into separate script which can be loaded via <svelte:head> to improve FOUC
- [ ] Add default themes (if not using Daisy, Skeleton, or custom provided) so works out of box without any customization
- [ ] Set `accent: hsl(var(--color-primary))` by default?
- [x] Refine `settings()` / `format()` integration (#191)
  - Mostly done, but needs to refine how locales are defined
- [ ] Write theme docs / migration guides, update customizations
- [ ] Re-read through https://github.com/techniq/svelte-ux/issues/66 and see if I missed something
- [ ] Determine if `initialVersions` be bumped to `0.58.0` for `svelte-ux`
- [ ] Publish and promote!

PR #171, opened 2024-01-03
fix: change overflow to auto, resolves #166

### LINKED ISSUE #166: [AppLayout] Change nav overflow from `scroll` to `auto`
See [discussion](https://discord.com/channels/920755200552226868/1191803609894105258)

Need to [change](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/AppLayout.svelte#L65-L67) `overflow-scroll` to `overflow-auto` so the scrollbars are only shown if the content overflows (more notable on Windows or Mac if you change your default scrollbar settings).

Semi related: #122 (but a bigger change, and will come after #66 is closed / #146 is merged).

thedivtagguy:
Simple fix for the scroll issue on AppLayout.

### DIFF
--- a/.changeset/wet-apricots-wait.md
+++ b/.changeset/wet-apricots-wait.md
@@ -1,5 +1,0 @@
----
-'svelte-ux': patch
----
-
-minor fix for Applayout
--- a/.changeset/wet-apricots-wait.md
+++ b/.changeset/wet-apricots-wait.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+minor fix for Applayout
--- a/.changeset/shy-peaches-ring.md
+++ b/.changeset/shy-peaches-ring.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+Update AppLayout nav to only show scrollbar when overflown (Windows, Mac with "Show scrollbars: always)
--- a/packages/svelte-ux/src/lib/components/AppLayout.svelte
+++ b/packages/svelte-ux/src/lib/components/AppLayout.svelte
@@ -62,7 +62,7 @@
       classes.aside
     )}
   >
-    <nav class={cls('nav h-full overflow-scroll w-[var(--navWidth)]', theme.nav, classes.nav)}>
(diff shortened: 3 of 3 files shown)


.changeset/wet-apricots-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+minor fix for Applayout
```
techniq, line 5: Could you change this to "Update AppLayout nav to only show scrollbar when overflown (Windows, Mac with "Show scrollbars: always)"

thedivtagguy: Pushed an updated changeset!

techniq:
Before | After
--- | ---
image | image

Note: The same will be true for the vertical scrollbar if there were not as many menu items to cause the overflow

On Mac, this is contingent on "Always" being checked (or having a non-Apple touch mouse / touchpad)

image

.changeset/wet-apricots-wait.md
techniq: Thanks.  You can actually edit the original file (`web-apricots-wait.md`).  Since you generated a new one, would you care to delete this file.

I need to have better docs for this process :(

Thanks!

thedivtagguy: Docs would be nice in the future, whenever time allows :) Hopefully I've done it right this time.

techniq:
@thedivtagguy Congrats on the first PR!  It will show up on the docs in the next 5 minutes (deploys currently take around 2-2.5 minutes, but sometimes they queue a few up).  I'll probably wait to include this in the release until we merge #167, which includes #137 and #146, but should land in the next 2 weeks (hopefully sooner).  Your changeset will then show up in the [CHANGELOG](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/CHANGELOG.md) and based on the level (patch, minor, major) will drive what version is incremented (major.minor.patch).  Since we're pre-1.0 ATM, everything is either `minor` (major features or breaking changes) or `patch` (everything else).  After 1.0, major will be breaking changes (and possibly major features), minor will be features, and patch will be non-breaking fixes, better known as semantic versioning, or [semver](https://semver.org/) for short.

Issue #170, opened 2024-01-03
Settings changes


techniq:
- [x] Rename `settings({ theme: {} })` to `settings({ classes: {} })`
- [x] Use `settings({ currentTheme: 'themeName' })` to set the theme (see #66 / #146), specifically `ThemeButton` ATM
- [x] Update `<Settings>` component to use top-level object properties as props. So `<Settings settings={{ theme: '...', classes: {...}, format: {...} }}>` becomes `<Settings theme="..." classes={{ ... }} format={{ ... }}>`

techniq:
cc: @jycouet @dimfeld - Some thoughts related to work you've done, or considering doing :)

dimfeld:
I think the main issue I see here is interaction between the `theme` set here and the default "system" (`prefers-color-scheme`) setting for dark mode. That could be resolved by changing theme to be something like:

```ts
type ThemeSetting = string | { light: string; dark: string };
```

In this case, a simple string would effectively disable the default `prefers-color-scheme` detection, but when both a light and dark theme are selected then it could choose appropriately. Of course, in all cases a theme setting in localStorage would override this.

What do you think?

techniq:
Trying to think this through.  Ultimately the source of truth is the `data-theme` attribute on `<html>` (or lack there of) and is how the theme is activated via CSS.  localStorage / Svelte store is only for the setting to remain between refreshes (and to actually update the DOM when changing).  It does need to [initialize](https://github.com/techniq/svelte-ux/blob/0b67ce5ce4c568e834a5df9ce0c279401f096319/packages/svelte-ux/src/lib/components/ThemeButton.svelte#L23-L29) the value, which is maybe the challenge with SSR we're talking about.

Last wins (if `data-theme` is set, it will be used (3), else if `prefers-color-scheme :dark` is true, it will be used (2), all else, the `:root` (aka `light`) will be used.  It's why the light theme is typically registered as both the default (`:root`) and explicitly using `[data-theme=light]`
![image](https://github.com/techniq/svelte-ux/assets/177476/95a8c090-eebd-42f4-9278-6ce28fc58711)

https://github.com/techniq/svelte-ux/blob/0b67ce5ce4c568e834a5df9ce0c279401f096319/packages/svelte-ux/src/lib/plugins/tailwind/theme.cjs#L117-L119

The themes should be registered in the tailwind config as...

```js
{
  ux: {
    themes: {
      light: { ... }, // will be `:root` and `[data-theme=light]`
[...]

PR #172, opened 2024-01-03
Change "theme" to "classes" in Settings

dimfeld:
- Rename settings `theme` argument to `classes`
- Rename getTheme to getClasses and getComponentTheme to getComponentClasses

Related to #170


techniq:
LGTM, and Vercel built without issue which is probably the best way to find any oddities.  Thanks!

PR #173, opened 2024-01-03
Add svelte settings to .prettierrc

dimfeld:
It was probably working for you since you're in VSCode which handles Prettier rather differently than the normal way, but this makes it work in other editors.


techniq:
Thanks

PR #174, opened 2024-01-03
Theme store

dimfeld:
This moves all the theme management into a store that lives outside any component, to allow multiple ways to set the theme.

The theme store lives in the Settings context. This also allows you to pass the list of themes into `settings`.

This does not cover SSR compatibility, which will come in a later PR

### DIFF
--- a/.changeset/modern-pumpkins-march.md
+++ b/.changeset/modern-pumpkins-march.md
@@ -1,0 +1,5 @@
+---
+"svelte-ux": patch
+---
+
+Add a store to manage the current theme
--- a/packages/svelte-ux/src/lib/components/MenuItem.svelte
+++ b/packages/svelte-ux/src/lib/components/MenuItem.svelte
@@ -28,7 +28,7 @@
   setButtonGroup(undefined);
 
   // Clear theme to not expose to Button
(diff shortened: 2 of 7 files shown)


techniq:
Quick read through and tested on the Vercel preview and seems to work as before, thanks!

One thing I noticed is if you are using system color scheme and change the emulation without refreshing, `class="dark"` doesn't reactively update, but it didn't before as well (and a case no one will really ever hit other than us likely).  Just an observation :)

https://github.com/techniq/svelte-ux/assets/177476/a917a784-fec7-4a24-9583-f97dabb3ff7e

It looks like the PR might need to be rebased from some stacked PRs.  I could probably merge fine, but if not too difficult, could you rebase?

dimfeld:
> One thing I noticed is if you are using system color scheme and change the emulation without refreshing, class="dark" doesn't reactively update, but it didn't before as well (and a case no one will really ever hit other than us likely). Just an observation :)

That's pretty easy to add though, and useful for debugging

dimfeld:
Done! Also rebased

techniq:
Thanks, Nice work!

PR #176, opened 2024-01-04
add ThemeInit component that place script inside <head> tag

dimfeld:
Needs testing but this should do it


dimfeld:
Ok verified this is working! Must be merged after #174 since it relies on it despite this particular branch not being based on it.

PR #177, opened 2024-01-04
Update Settings component to expose each settings value as its own component prop

### LINKED ISSUE #170: Settings changes

dimfeld:
Resolves #170

### DIFF
--- a/.changeset/chilled-dingos-sell.md
+++ b/.changeset/chilled-dingos-sell.md
@@ -1,0 +1,5 @@
+---
+"svelte-ux": minor
+---
+
+Update Settings component to expose each settings key as a component prop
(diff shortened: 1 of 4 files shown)


techniq (approved):
Slick use of $$restProps.  Thanks!

Issue #163, opened 2023-12-31
[TextField] Can not enter decimal separator (Safari)


pekeler:
When entering a `.` or `,` as in `12.34`, the value gets deleted.

pekeler:
(This is in Safari 17.1, seems to work fine in Chrome)

techniq:
Thanks for the report @pekeler.  Yeah, it appears to only affect Safari (also tested Firefox).  

My first thought was how Safari might he handling `Number('12.')` but it looks consistent with Chrome.

Chrome | Safari
--- | ---
![image](https://github.com/techniq/svelte-ux/assets/177476/38f74d12-2729-416b-b2cb-d1358abf8a77) | ![image](https://github.com/techniq/svelte-ux/assets/177476/663b32af-9493-40fa-aa0a-1374a5167d1e)

I've had a note to [improve](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/TextField.svelte#L127-L150) this functionality.  Hopefully I (or someone else) can get to it soon.  The current workaround is not using `type="decimal"` and manually converting the value.

I do wonder if some of this handling is needed, or shouldn't be replaced with `e.target.valueAsNumber` (although Svelte already does this for `<input type="number" bind:value >`.

dimfeld:
Hmm, this doesn't reproduce for me, using Safari 17.2.1. I just tried it with the sample decimal TextField in the docs site.

dimfeld:
Posted this on the Discord, but here for the record. This appears to be a bug in Safari 17.1, fixed in 17.2
![image](https://github.com/techniq/svelte-ux/assets/1369558/4bea068e-4780-43c0-9a7a-f933b33dec29)

techniq:
> not returning the correct value

You could say that :) 

![image](https://github.com/techniq/svelte-ux/assets/177476/91e2a63a-fcae-4346-92ab-1f45e7e49243)

techniq:
@pekeler I'm going to close this as I don't see a way to workaround it otherwise, and at least it's resolved in the latest Safari release :)

pekeler:
Makes sense. Thanks.

(Safari is my default browser. Quite interesting I never noticed this browser bug on any other site. I guess almost nobody is using real number inputs and does their own parsing.)

157Thor:
Can we reopen this? This issue is happening for me on chrome and brave

PR #178, opened 2024-01-04
Fix cursor jumping when backspacing near a decimal point in a decimal TextField

### LINKED ISSUE #21: TextField number type (integer, decimal, ...) improvements
- [x] Improve handling of empty string (parsed to `0` but can backspace out (but still shows value instead of shrunk label / placeholder))
- [ ] If value `1234.05` set and backspace `5`, `1234` is immediately set (removes `.0`) due to `Number(1234.0)` resulting in `1234`

dimfeld:
Fixes #21 
Might also address #163 but I couldn't reproduce that one on my end.

### DIFF
--- a/.changeset/shiny-hornets-study.md
+++ b/.changeset/shiny-hornets-study.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+Prevent cursor jumping around when backspacing near a decimal point in a number TextField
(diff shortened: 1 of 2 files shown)


techniq:
Thanks!  Quick peak and it doesn't look to fix the Safar bug (#163) 

https://github.com/techniq/svelte-ux/assets/177476/2c74a274-6271-4cd3-b0b6-444d2f5c7ef2

techniq (approved):
LGTM.  Thanks!

PR #179, opened 2024-01-04
Add zero-width-joiner character in import.meta.env to prevent spurious warning from svelte-kit package

dimfeld:
Run the package command and note the lack of the warning :)

This does show up as `<200d>` in Vim, but it's invisible for VS Code so I think that's ok.


techniq:
Awe, this one :) 

image

Nice workaround :).  Transparent in the PR diff, but I'll take your word for it.  Thanks!

techniq (approved):

dimfeld:
CleanShot 2024-01-04 at 15 10 48

☝️ in neovim, kind of ugly but its ok us vim users are a relatively small group

PR #191, opened 2024-01-07
Move format function definitions onto settings to avoid calling getSettings after component initialization

dimfeld:


dimfeld:
This is pretty much done. Just need to clean up some of the unit tests.

techniq (approved):
LGTM.  Thanks @dimfeld

PR #202, opened 2024-01-09
Locale support cleanup

dimfeld:


.changeset/sixty-walls-rest.md
```diff
@@ -0,0 +1,6 @@
+---
+"svelte-ux": minor
+---
+
+- Add locale settings to settings object, and allow dynamically changing the current locale.
+- Move format function to be a store on the settings object, which updates when the locale changes.
```
techniq, line 6: Is this true?

techniq: Thanks.  I was making sure I hadn't overlooked or misunderstood something.  We [still](https://github.com/techniq/svelte-ux/blob/fdeb597872bae0fcd99273782572d72eb62d039f/packages/svelte-ux/src/lib/utils/format.ts#L17-L37) have `format()` as a non-store function as well, which I'd like to keep that way for the time being.

techniq (approved):
LGTM.  Thanks!

PR #203, opened 2024-01-09
Format last qol for next

jycouet:
- [ ] start of `<LanguageSelect />`
- [x] add default `weekStartOn` depending on local
- [x] updated tests
- [x] fix some types

_I'll add some comment in the code_

### DIFF
--- a/packages/svelte-ux/src/lib/components/LanguageSelect.svelte
+++ b/packages/svelte-ux/src/lib/components/LanguageSelect.svelte
@@ -1,0 +1,49 @@
+<script lang="ts">
+  import Button from './Button.svelte';
+  import Menu from './Menu.svelte';
+  import MenuItem from './MenuItem.svelte';
+  import { cls } from '../utils/styles';
+  import { getSettings } from './settings';
+  const { locale } = getSettings();
+
+  let open = false;
+
+  type Language = {
+    name: string;
+    code: string;
+    flag: string;
+  };
+
+  export let languagesDemo: Language[] = [
+    { name: 'English', code: 'en', flag: '🇺🇸' },
+    { name: 'Français', code: 'fr', flag: '🇫🇷' },
+    // add more for the demo
+  ];
+  $: languageSelected = languagesDemo.find((c) => c.code === $locale)!;
+</script>
+
+<Button on:click={() => (open = !open)}>
+  {languageSelected.flag}
+  <Menu bind:open on:close={() => (open = false)} offset={4} explicitClose resize>
+    <div class="grid gap-2 p-2 border-b border-surface-content/10">
+      {#each languagesDemo as language}
+        <MenuItem
+          on:click={() => {
+            languageSelected = language;
+            locale.set(language.code);
+          }}
+          class={cls(
+            'bg-surface-100 text-surface-content font-semibold border shadow',
+            languageSelected === language && 'ring-2 ring-surface-content'
+          )}
+        >
+          {language.flag} - {language.name}
+        </MenuItem>
+      {/each}
+    </div>
+
+    <div class="p-2 grid grid-cols-[auto,1fr] gap-2 items-center text-xs">
+      <span class="font-medium">Affect dates & numbers formats</span>
+    </div>
+  </Menu>
+</Button>
--- a/packages/svelte-ux/src/lib/utils/date.test.ts
+++ b/packages/svelte-ux/src/lib/utils/date.test.ts
@@ -1,25 +1,28 @@
 import { describe, it, expect } from 'vitest';
 import {
-  PeriodType,
   formatDate,
   getMonthDaysByWeek,
   localToUtcDate,
   utcToLocalDate,
-  DayOfWeek,
   formatIntl,
-  type CustomIntlDateTimeFormatOptions,
-  type FormatDateOptions,
-  DateToken,
   formatDateWithLocale,
 } from './date';
-import { getSettings } from '$lib/components';
-import { format, formatWithLocale } from '.';
+import { formatWithLocale } from '.';
 import { createLocaleSettings, defaultLocale } from './locale';
+import {
+  PeriodType,
+  type FormatDateOptions,
+  DayOfWeek,
+  type CustomIntlDateTimeFormatOptions,
+  DateToken,
+} from './date_types';
+import { getWeekStartsOnFromIntl } from './dateInternal';
 
 const DATE = '2023-11-21'; // "good" default date as the day (21) is bigger than 12 (number of months). And november is a good month1 (because why not?)
 const dt_2M_2d = new Date(2023, 10, 21);
 const dt_2M_1d = new Date(2023, 10, 7);
 const dt_1M_1d = new Date(2023, 2, 7);
+const dt_first = new Date(2024, 1, 1);
 
 const dt_1M_1d_time_pm = new Date(2023, 2, 7, 14, 2, 3, 4);
 const dt_1M_1d_time_am = new Date(2023, 2, 7, 1, 2, 3, 4);
@@ -30,9 +33,6 @@ const fr = createLocaleSettings({
     dates: {
       ordinalSuffixes: {
         one: 'er',
-        two: '',
-        few: '',
-        other: '',
       },
     },
   },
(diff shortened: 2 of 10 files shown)


packages/svelte-ux/src/lib/utils/dateInternal.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function getWeekStartsOnFromIntl(locales?: string): DayOfWeek {
+  if (!locales) {
+    return DayOfWeek.Sunday;
+  }
+
+  const info = new Intl.Locale(locales);
+  // @ts-ignore
+  return (info.weekInfo.firstDay ?? 0) % 7; // (in Intl, sunday is 7 not 0, so we need to mod 7)
+}
```
jycouet, line 11: I don't know why, but if I put this function in `date.ts`, I can't call it from `locale.ts`!

I have a message like: `getWeekStartsOnFromIntl is not a function.`

packages/svelte-ux/src/lib/components/LanguageSelect.svelte
```diff
@@ -0,0 +1,49 @@
...
+
+<Button on:click={() => (open = !open)}>
+  {languageSelected.flag}
+  <Menu bind:open on:close={() => (open = false)} offset={4} explicitClose resize>
+    <div class="grid gap-2 p-2 border-b border-surface-content/10">
+      {#each languagesDemo as language}
+        <MenuItem
+          on:click={() => {
+            languageSelected = language;
+            locale.set(language.code);
```
jycouet, line 33: This is changing the `locale`, but when I'm on the DatePicker I don't see any change.

packages/svelte-ux/src/routes/+layout.svelte
```diff
@@ -32,61 +33,32 @@
...
-    //     },
-    //   },
-    // },
-    // dictionary: {
-    //   Cancel: 'Annuler',
-    //   Date: {
-    //     Day: 'Jour',
-    //   },
-    // },
+    // fallbackLocale: 'fr',
```
jycouet, line 36: When I set `fallbackLocale: 'fr'` all dates seems to work well 👍

packages/svelte-ux/src/routes/+layout.svelte
```diff
@@ -32,61 +33,32 @@
...
+            ordinalSuffixes: {
+              one: 'er',
+            },
+          },
+          numbers: {
+            defaults: {
+              currency: 'EUR',
+            },
+          },
+        },
```
jycouet, line 52: So now `locale` is for dates & numbers at the same time? For my use case it's fine, just want to be sure it's intentional.

At the same time, we repeat 2 times "fr". Maybe `localesFormats` should be an array?

packages/svelte-ux/src/lib/utils/date.test.ts
```diff
@@ -138,32 +141,28 @@ describe('formatDate()', () => {
...
-      [PeriodType.Week, 'long', fr, DayOfWeek.Monday, '20/11/2023 - 26/11/2023'],
+      [PeriodType.Week, 'short', defaultLocale, '11/19 - 11/25'],
+      [PeriodType.Week, 'short', fr, '20/11 - 26/11'],
+      [PeriodType.Week, 'long', defaultLocale, '11/19/2023 - 11/25/2023'],
+      [PeriodType.Week, 'long', fr, '20/11/2023 - 26/11/2023'],
+
+      [PeriodType.Week, 'short', defaultLocale, '11/19 - 11/25'],
+      [PeriodType.Week, 'short', fr, '20/11 - 26/11'],
+      [PeriodType.Week, 'long', defaultLocale, '11/19/2023 - 11/25/2023'],
+      [PeriodType.Week, 'long', fr, '20/11/2023 - 26/11/2023'],
```
jycouet, line 159: Updates test to show Week with automatic start of the week. 🎉

packages/svelte-ux/src/lib/components/LanguageSelect.svelte
```diff
@@ -0,0 +1,49 @@
...
+
+  export let languagesDemo: Language[] = [
+    { name: 'English', code: 'en', flag: '🇺🇸' },
+    { name: 'Français', code: 'fr', flag: '🇫🇷' },
+    // add more for the demo
+  ];
+  $: languageSelected = languagesDemo.find((c) => c.code === $locale)!;
+</script>
+
+<Button on:click={() => (open = !open)}>
```
techniq, line 25: [MenuButton](https://format-last-qol-for-next.svelte-ux.pages.dev/docs/components/MenuButton) could simplify this a good bit, although we'd probably need to access a slot for the note about affecting dates and numbers like the `actions` slot of `SelectField`).  Probably not worth changing now, just mentioning it if you hadn't seen it.  The `stepper` functionality` is rather nice (and I've added `stepper` to `SelectField` in my current theme generator PR)

packages/svelte-ux/src/lib/components/LanguageSelect.svelte
techniq: hmm.. I noticed this as well (same for the other Date* components)..  Have you looked into it yet?

packages/svelte-ux/src/lib/utils/date.test.ts
```diff
@@ -30,21 +33,21 @@ const fr = createLocaleSettings({
...
-        few: '',
-        other: '',
       },
     },
   },
 });
 
 describe('formatDate()', () => {
   it('should return empty string for null or undefined date', () => {
+    // @ts-expect-error
```
techniq, line 43: Should this error?

packages/svelte-ux/src/lib/utils/date.test.ts
techniq: Nice!

packages/svelte-ux/src/lib/utils/dateInternal.ts
techniq: hmm, that is odd.  I'll try to look at it locally, but we can merge as is for now.

packages/svelte-ux/src/routes/+layout.svelte
techniq: If you're fine with the single locale, I definitely am.  I'm not aware of a use case of where we would need to have them separate, especially from a UI/selection standpoint.  There would have to be a really strong reason to support individual locales per date/number/etc :) 

The object makes it easier/faster to lookup the value instead of `arr.find(...)`.

packages/svelte-ux/src/routes/+layout.svelte
```diff
@@ -191,7 +163,8 @@
 
       <QuickSearch options={quickSearchOptions} on:change={(e) => goto(e.detail.value)} />
 
-      <div class="border-r border-primary-content/20 pr-2 grid items-center">
+      <div class="border-r border-primary-content/20 pr-2 grid grid-cols-2 items-center">
+        <LanguageSelect />
```
techniq, line 167: YAY!  The beginning of showing a lot of this hard work in the docs (especially how the Date* components are not Intl.

I know we need to figure out the reactivity / setting issue, and UI wise I'll refine this before we publish to production (especially since we don't have room for another icon on mobile, and already need to reduce what we have.

techniq (approved):
Quick review while I had a moment.  Nothing major jumping out at me other than the reactivity issue.  We can still merge with the issue and troubleshoot after words (will make a followup PR by any of us easier that way).

I left a few comments, just let me know if there is anything else you would want to handle before I merge.

packages/svelte-ux/src/routes/+layout.svelte
jycouet: Switzerland is a good example of complicated locales ^^ with some french, some german, currency in CHF... 😅

But yes nvm for now, and it can evolve, so all good.

packages/svelte-ux/src/lib/components/LanguageSelect.svelte
jycouet: Thx for the info. 
I copy / pasted the `<ThemeSelect />` as a starting point, that's why ^^

packages/svelte-ux/src/lib/utils/date.test.ts
jycouet: Yes, because the number of args doesn't match anymore.

packages/svelte-ux/src/routes/+layout.svelte
jycouet: Yep, I leave you the hand now :)

PR #208, opened 2024-01-10
Feat/date range weekstarton

jycouet:


packages/svelte-ux/src/lib/utils/dateRange.ts
```diff
@@ -1,35 +1,47 @@
...
   periodType?: PeriodType | null;
 };
 
-export const dayPresets = getDateRangePresets(PeriodType.Day);
-export const biWeekSun1Presets = getDateRangePresets(PeriodType.BiWeek1Sun);
-export const biWeekMon1Presets = getDateRangePresets(PeriodType.BiWeek1Mon);
-export const monthPresets = getDateRangePresets(PeriodType.Month);
-export const quarterPresets = getDateRangePresets(PeriodType.Quarter);
-export const fiscalYearPresets = getDateRangePresets(PeriodType.FiscalYearOctober);
+// All this is exported, but never used. Is it usefull?
```
techniq, line 13: I thought I used these for something, but I checked a few of my other projects and didn't find any.  We could probably remove the `export`

packages/svelte-ux/src/routes/+layout.svelte
```diff
@@ -52,6 +52,16 @@
           Cancel: 'Annuler',
```
techniq, line 52: Might be nice to go ahead and add `OK` translation since we already have Cancel, and start to sprinkle it about...

image

image

packages/svelte-ux/src/routes/docs/components/DateRange/+page.svelte
```diff
@@ -57,8 +60,8 @@
 <Preview>
   <DateRange
     periodTypes={[PeriodType.Day, PeriodType.Month]}
-    getPeriodTypePresets={(periodType) => {
-      const { start, end, add } = getDateFuncsByPeriodType(periodType);
+    getPeriodTypePresets={(settings, periodType) => {
```
techniq, line 63: This should be `$localeSettings`

packages/svelte-ux/src/routes/+layout.svelte
```diff
@@ -31,7 +31,7 @@
   $: title = data.pr_id ? `🚧 (pr:${data.pr_id}) - ${baseTitle}` : baseTitle;
 
   settings({
-    // fallbackLocale: 'fr',
+    fallbackLocale: 'fr',
```
techniq, line 34: Don't forget to remove this when you're ready to merge :)

techniq: Maybe we consider storying settings in `localStorage` so their persisted between sessions.  It's a bit of pain to maintain that contract of the persisted model (we'd just need to validate and maybe version).

Worst case we could just persist the `fallbackLocale` in the docs to localStorage (simple string)

packages/svelte-ux/src/lib/utils/date.ts
```diff
@@ -57,15 +57,27 @@ export function getPeriodTypeName(periodType: PeriodType) {
   return getPeriodTypeNameWithLocale(defaultLocale, periodType);
 }
 
+// util to make sure we have handled all enum cases
+function assertNever(x: never): never {
+  throw new Error(`Unhandled enum case: ${x}`);
+}
+
```
techniq, line 64: Should we move this to `typeHelpers`?  I know the error message isn't exactly generic though...

techniq (commented):
Looking good so far

packages/svelte-ux/src/routes/+layout.svelte
jycouet: Ok, I gave a shot at it. Let's see

packages/svelte-ux/src/routes/docs/components/DateRange/+page.svelte
jycouet: I think that it's name of args, not `$localeSettings`, but I'm maybe wrong

packages/svelte-ux/src/routes/+layout.svelte
jycouet: Maybe yes something to do (fallbackLocale)

packages/svelte-ux/src/lib/utils/date.ts
jycouet: It's not really about the error message, it's more that when we will be able to have `pnpm check`, type will complain at the `switch` level and showing what case is not handled.

I move it to `typeHelpers` and added a bit of doc on top

packages/svelte-ux/src/lib/utils/dictionary.ts
```diff
@@ -3,16 +3,37 @@ export type DictionaryMessagesOptions = {
...
+    PeriodeFiscalYear?: PeriodeDayMsg;
   };
 };
 
+export type PeriodeDayMsg = {
+  Current?: string;
+  Last?: string;
+  LastX?: string;
+};
+
```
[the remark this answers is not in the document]
jycouet: Oups, all the way to french haha, life is hard ^^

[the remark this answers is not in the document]
techniq: I'll rename `dico` later :)

PR #214, opened 2024-01-13
Handle undefined week info

dimfeld:
This fixes the crashes in Safari

### DIFF
--- a/packages/svelte-ux/src/lib/utils/dateInternal.ts
+++ b/packages/svelte-ux/src/lib/utils/dateInternal.ts
@@ -7,5 +7,5 @@ export function getWeekStartsOnFromIntl(locales?: string): DayOfWeek {
 
   const info = new Intl.Locale(locales);
   // @ts-ignore
-  return (info.weekInfo.firstDay ?? 0) % 7; // (in Intl, sunday is 7 not 0, so we need to mod 7)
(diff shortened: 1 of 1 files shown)


techniq:
Yay web standards..

Chrome | Safari
--- | ---
image | image

based on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo) / [CanIUse](https://caniuse.com/?search=intl.getweekinfo),  `getWeekInfo()` is the correct method, but Chrome/Edge/etc uses `weekInfo` (and maybe update to `getWeekInfo()` in the future.

Can we change...

```js
return (info.weekInfo?.firstDay ?? 0) % 7;
```
to...
```js
return ((info.weekInfo ?? info.getWeekInfo?.())?.firstDay ?? 0) % 7
```
or maybe...
```js
const weekInfo = info.weekInfo ?? info.getWeekInfo?.();
return (weekInfo?.firstDay ?? 0) % 7
```

Oh, and THANKS! for tracking down the root cause for the Safari error: `Unhandled Promise Rejection: ReferenceError: Cannot access uninitialized variable.`

cc: @jycouet

jycouet:
Yeahhh ☺️
Web std is great but...

Nice fix 👍

PR #215, opened 2024-01-13
update SvelteKit and Vite dependencies

dimfeld:


techniq:
Thanks!

techniq:
@dimfeld Btw, I think my main issue with `fsevents` / etc is related to trying to get `vitest` updated to `1.x` which I just noticed this PR didn't do (I initially thought it did, and assumed Vite 5 might require this, but it doesn't).  I still want to get onto the latest Vite 5 / SvelteKit 2, but I would love if you could take a look at updating vitest and see if you get similar issues.

One reason we stayed on `0.33` instead of going to `0.34` was due to an:

> Failed to resolve entry for package "@sveltejs/package"

but I think adding `@sveltejs/vite-plugin-svelte` is the fix for that error (see this [comment](https://github.com/techniq/svelte-ux/pull/187#issuecomment-1884080929))

One thing I just noticed is the automated [PR](https://github.com/techniq/svelte-ux/pull/187#issuecomment-1884080929) to update vitest to `0.34` is actually failing CI due to `fsevents`, so definitely `vitest` related 🤔

PR #219, opened 2024-01-15
Upgrade vitest to 1.2

### LINKED ISSUE #216: Update vitest to 1.x
We are currently on `vitest` `0.33` as we ran into `Failed to resolve entry for package "@sveltejs/package"` when updating to `0.34` initially.  This sounds like it can be solved by adding `@sveltejs/vite-plugin-svelte` as a direct dependency.

I recently tried to update to the latest `1.2.0` as part of updating to Vite 5 / SvelteKit 2, and ran into `fsevent` errors.  See CI [logs](https://github.com/techniq/svelte-ux/actions/runs/7495881658/job/20406823925?pr=187) on the automated [PR](https://github.com/techniq/svelte-ux/pull/187#issuecomment-1884080929) to update to `0.34`

dimfeld:
The problem with fsevents was due to dependency prebundling of sveld, which isn't used in the tests but was slurped up in the prebundle process. The trick here was that it had to be disabled in the *test* configuration, not in the main Vite configuration.

Resolves #216

### DIFF
--- a/packages/svelte-ux/package.json
+++ b/packages/svelte-ux/package.json
@@ -33,7 +33,7 @@
     "@types/lodash-es": "^4.17.12",
     "@types/marked": "^6.0.0",
     "@types/prismjs": "^1.26.3",
-    "@vitest/coverage-v8": "^0.34.6",
+    "@vitest/coverage-v8": "^1.2.0",
     "autoprefixer": "^10.4.16",
     "daisyui": "^4.6.0",
     "execa": "^8.0.1",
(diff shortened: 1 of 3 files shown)


techniq:
Awesome, thanks so much!  I might move `sveld` into a `@svelte-ux/docs` package (with some other items) so we can move it out of `dependencies` (mostly for `LayerChart`) and add `@svelte-ux/docs` in `devDependencies` for both Svelte UX and LayerChart.

PR #228, opened 2024-02-03
Fix FOUC with ThemeSwitch in SSR

### LINKED ISSUE #227: next branch: ThemeSwitch FOUC with SSR
ThemeSwitch shows the light mode setting then moves over to dark mode if loading in SSR. This is because the positioning is just using the checked state instead of being keyed on the `dark` class.

This is a GIF of me refreshing the page three times.

CleanShot 2024-02-02 at 21 07 12

This can probably just be overridden by passing in some custom classes to disable the normal positioning classes and use `dark:` variant based code instead. I'll make a PR for this soon-ish.

dimfeld:
Closes #227

### DIFF
--- a/.changeset/nice-cats-know.md
+++ b/.changeset/nice-cats-know.md
@@ -1,0 +1,5 @@
+---
+"svelte-ux": patch
+---
+
+Fix ThemeSwitch SSR behavior by using `dark` class to define positioning and icon display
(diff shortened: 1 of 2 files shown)


techniq:
Thanks @dimfeld.   I plan to add a `variant` to `ThemeSwitch` that allows it to be rendered as a single visible icon that has a transition as well (similar to your original SSR version).

PR #230, opened 2024-02-06
Add defaultProps to settings, with labelPlacement setting

dimfeld:
Probably needs some update to the docs site to demonstrate this.


packages/svelte-ux/src/lib/components/theme.ts
```diff
@@ -1,25 +1,42 @@
...
     ? ComponentProps<PR>['classes']
     : never
   : never;
 
+export type ResolvedComponentClasses = {
+  [key in ComponentName]: ClassesProp<(typeof Components)[key]> extends never
+    ? {}
+    : NonNullable<ClassesProp<(typeof Components)[key]>>;
+};
+
```
dimfeld, line 18: TS wasn't doing a good job of inferring the types, giving all sorts of weird errors once I tried to factor out resolveComponentClasses. This allowed me to just explicitly name the type, with the added benefit of full type info on the returned settingsClasses in each component.

dimfeld:
@techniq Might make one last change here so that instead of returning `{ classes, ...defaultProps }` it returns `{ settingsClasses, defaultProps }` which should make it easier for the wrapper components to do

```
<script>
  const { settingsClasses, defaultProps } = getComponentSetttings('MultiSelectField');

  $: fieldRestProps = { ...$$restProps, ...defaults };
</script>
<Field {...fieldRestProps} />
```

Should ease the burden of adding additional defaults in the future since they won't have to be explicitly accounted for in every component wrapping Field, Button, etc..

dimfeld:
Did the above. Improves the ergonomics there and allows me to remove a `ts-expect-error` :)

techniq:
@dimfeld I noticed the AppBar styling was lost, but it was a simple [fix](https://github.com/techniq/svelte-ux/pull/230/commits/a3a3dfa9a4a72174d4875e39cffa01283f9c109c).

image

image

packages/svelte-ux/src/lib/components/settings.ts
```diff
@@ -111,6 +125,37 @@ export function getSettings(): Settings {
...
+
+  return output;
+}
+
+export function getComponentSettings<NAME extends ComponentName>(
+  name: NAME
+): ResolvedComponentSettings<NAME> {
+  const settings = getSettings();
+
+  const existing = settings.componentSettingsCache[name];
```
techniq, line 153: Using a cache will mean we can't change these at runtime, but not a big use case other than docs / demos.  The performance improvements (likely small, but add up being in the hot path) are worth it I think.

techniq:
@dimfeld Looks great!  I pushed a few fixes/changes to your branch. 

- Fix `getComponentClasses()` for string classes (root-only).  See above screenshots
- Added `color` and `size` settings() for Button.  Seems helpful, especially if also setting `variant`
- Moved all the types from `types/options.ts` to `types/index.ts`
- Tweaked the customization docs a little

One small comment regarding using the cache (I think it's worth it, just a consideration), and whether we want to hard code some `default` Button variants/etc in places.

One place that comes to mind is on dialogs like `DateRangeField` and `MultiSelectField`

DateRangeField | MultiSelectField
--- | ---
image | image

I could see supporting a `primaryButtonVariant` and `secondaryButtonVariant` (or maybe `applyButtonVariant` and `cancelButtonVariant` but there might come a time when you have button actions that are not OK/Cancel, or have more than 2 (none come to mind in Svelte UX though)

These screenshots also highlight that I need to be consistent with button ordering for these (been on my mind for a while).  I've considered exposing that as a setting as some point as well.  Comes down to Windows/Mac and preference.   I've slowly changed my mind that the primary should be in the bottom right corner like Mac for dialogs, but reading left/right and things like drawers feel better with the primary on the left.

Windows | Mac
--- | ---
image | image

Anyways, more context than we need for this PR.  I'm fine with merging as is now (looks great!) and we refine/solve these later.

Thanks again!

packages/svelte-ux/src/lib/components/settings.ts
dimfeld: Yeah you can't change them at runtime anyway since it's not a store. At least not without reaching directly into the object and then recreating any controls that you want to change, and in that case you might as well also just set the cache back to `{}` :)

When you call `settings` again it does create a new cache for that new instance of the settings object, so you can still do "sub-settings" like you do in a couple places.

dimfeld:
> * Fix `getComponentClasses()` for string classes (root-only).  See above screenshots

Ah this is actually not correct since getComponentSettings passes `classes` directly, and so getComponentClasses needs to change to do the same. Then all the references in getComponentClasses should just be `classes.`.

 Wish TS would have caught this, I guess since `classes` is sort of like a Record it didn't.

Anyway, quick change, I just updated it

packages/svelte-ux/src/lib/components/settings.ts
techniq: Awe, good point.  Hadn't really though tit through.  Thanks!

techniq:
> > * Fix `getComponentClasses()` for string classes (root-only).  See above screenshots
> 
> Ah this is actually not correct since getComponentSettings passes `classes` directly, and so getComponentClasses needs to change to do the same. Then all the references in getComponentClasses should just be `classes.`.
> 
> Wish TS would have caught this, I guess since `classes` is sort of like a Record it didn't.
> 
> Anyway, quick change, I just updated it

Awe, good catch.  Thanks!  I'm surprised I didn't notice an issue while playing around with it locally, but glad you caught it.

PR #244, opened 2024-02-09
feat(DateField): Add support for `class` and `classes` props

brandonmcconnell:
This PR adds support for `class` and `classes` props on DateField, and passes a `DateField` class name which can be used to more specifically target DateFields if/when necessary

### DIFF
--- a/packages/svelte-ux/src/lib/components/DateField.svelte
+++ b/packages/svelte-ux/src/lib/components/DateField.svelte
@@ -22,11 +22,9 @@
   $: actualFormat = format ?? $format_ux.settings.formats.dates.baseParsing ?? 'MM/dd/yyyy';
   $: actualMask = mask ?? actualFormat.toLowerCase();
 
-  type FieldClasses = NonNullable<ComponentProps<Field>['classes']>;
-
   export let classes: {
-    root?: FieldClasses['root'];
-    field?: FieldClasses;
+    root?: string;
+    field?: ComponentProps<Field>['classes'];
   } = {};
 
   // Field props
--- a/packages/svelte-ux/src/lib/components/DateField.svelte
+++ b/packages/svelte-ux/src/lib/components/DateField.svelte
@@ -1,8 +1,9 @@
 <script lang="ts">
-  import { createEventDispatcher } from 'svelte';
+  import { createEventDispatcher, type ComponentProps } from 'svelte';
   import { parse as parseDate } from 'date-fns';
   import { PeriodType } from '../utils';
   import { getComponentSettings, getSettings } from './settings';
+  import { cls } from '../utils/styles';
 
   import Field from './Field.svelte';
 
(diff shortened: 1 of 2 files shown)


packages/svelte-ux/src/lib/components/DateField.svelte
```diff
@@ -65,6 +78,15 @@
...
   }}
+  classes={{
+    root: cls('DateField', settingsClasses.root, classes.root, $$props.class),
+    container: cls(settingsClasses.container, classes.container),
+    label: cls(settingsClasses.label, classes.label),
+    input: cls(settingsClasses.input, classes.input),
+    error: cls(settingsClasses.error, classes.error),
+    prepend: cls(settingsClasses.prepend, classes.prepend),
+    append: cls(settingsClasses.append, classes.append),
+  }}
```
brandonmcconnell, line 89: btw with something like a `clsMulti` (bikeshed that some hah), this could be reduced to…

```jsx
classes={clsMulti(
  settingsClasses,
  classes,
  { root: cls('DateField', $$props.class) }
)}
```

packages/svelte-ux/src/lib/components/DateField.svelte
```diff
@@ -21,6 +22,13 @@
   $: actualFormat = format ?? $format_ux.settings.formats.dates.baseParsing ?? 'MM/dd/yyyy';
   $: actualMask = mask ?? actualFormat.toLowerCase();
 
+  type FieldClasses = NonNullable<ComponentProps<Field>['classes']>;
```
techniq, line 25: Is `NonNullable` doing anything here?

packages/svelte-ux/src/lib/components/DateField.svelte
```diff
@@ -21,6 +22,13 @@
...
+
+  export let classes: {
+    root?: FieldClasses['root'];
```
techniq, line 28: This can just be `string` in this case

techniq (changes requested):
Few small comments

packages/svelte-ux/src/lib/components/DateField.svelte
brandonmcconnell: `NonNullable` was only being used to more easily access the optimistic type for `root`. I considered we might just want `string` but wasn't sure if the type should be derived.

I'll replace this.

packages/svelte-ux/src/lib/components/DateField.svelte
brandonmcconnell: Yep! I'll replace that and remove `NonNullable`

packages/svelte-ux/src/lib/components/DateField.svelte
brandonmcconnell: Resolved in [209b2c3](https://github.com/techniq/svelte-ux/pull/244/commits/209b2c393bdfd142f1b5e902f9b618350ffaf7bb)

packages/svelte-ux/src/lib/components/DateField.svelte
brandonmcconnell: Resolved in [209b2c3](https://github.com/techniq/svelte-ux/pull/244/commits/209b2c393bdfd142f1b5e902f9b618350ffaf7bb)

techniq (approved):
Looks good.  Thanks!

PR #260, opened 2024-02-15
Place vertical padding on Input element instead of border when not using an inset label

dimfeld:
This almost gets it there... There's still a 1 pixel area where clicking doesn't work right on the border.


dimfeld:
Closing per further experimentation and discussion in Discord

PR #262, opened 2024-02-15
ToggleOption: Move click handlers from <label> to <input>

dimfeld:
This fixes an a11y warning. Browsers fire the click event from the `input` even when the `label` is clicked, so this works. 

We may also want some focus styling on this control to help with keyboard navigation.

Also added some TS type checking fixes.

### DIFF
--- a/packages/svelte-ux/src/lib/components/ToggleOption.svelte
+++ b/packages/svelte-ux/src/lib/components/ToggleOption.svelte
@@ -45,8 +45,6 @@
 
 <label
   class:selected
-  on:click={() => selectOption(optionElement, value)}
-  on:click
   bind:this={optionElement}
   {...$$restProps}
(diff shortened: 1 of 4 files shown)


dimfeld:
Something like `has-[:focus]:text-primary/70` seems to work ok, probably room for improvement though. Tabbing is a little weird too since these look like buttons (in ToggleGroup) but under the hood it's a radio button group.

techniq:
Thanks @dimfeld.  Curious how you saw the a11y warning.  I'm surprised VSCode / svelte-check wasn't complaining at me when I opened the current file (like it usually does).

image

dimfeld:
It was the first thing that popped up when I ran pnpm dev. Yeah I didn’t see it in the editor either, not sure why...

PR #264, opened 2024-02-16
Fix warnings about missing prop settingsClasses.root

dimfeld:
- Update ClassesProp to detect if a `class` string prop exists.
- Add explicit `class` prop where `$$props.class` was used before. This both informs the type system and improves autocomplete experience for users.


techniq:
Thanks @dimfeld.  As much as I dislike `export { className as class };`, I concede it needs to be done.  I wish `type $$RestProps` or similar was possible, but sadly it's not (or in the [rfc](https://github.com/dummdidumm/rfcs/blob/ts-typedefs-within-svelte-components/text/ts-typing-props-slots-events.md#typing-props))

Issue #271, opened 2024-02-25
vitePreprocess import error when you run pnpm run dev on apps created with pnpm create svelte-ux@latest


bamarb:
Hello 

I get the following error when i try to create a svelte-ux app. I have a  patch ready would be glad to contribute.
Steps to reproduce :

pnpm create svelte-ux@latext svux-fun

cd svux-fun
pnpm i 

pnpm run dev

> try-svux@0.0.1 dev /shared/scratch/svelte/try-svux
> vite dev

▲ [WARNING] Cannot find base config file "./.svelte-kit/tsconfig.json" [tsconfig.json]

    tsconfig.json:2:13:
      2 │   "extends": "./.svelte-kit/tsconfig.json",
        ╵              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

error when starting dev server:
file:///shared/scratch/svelte/try-svux/svelte.config.js?ts=1708839040624:2
import { vitePreprocess } from '@sveltejs/kit/vite';
         ^^^^^^^^^^^^^^
SyntaxError: The requested module '@sveltejs/kit/vite' does not provide an export named 'vitePreprocess'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)
 ELIFECYCLE  Command failed with exit code 1.

techniq:
Thanks for the report and PR @bamarb.  Fix released as `0.3.2` (latest)

techniq:
Fixed via https://github.com/techniq/svelte-ux/pull/272

PR #272, opened 2024-02-25
Fix vitePreprocess import error when app created with npm create svel…

bamarb:
Fix for issue #271


techniq (approved):
Thanks @bamarb.  I appreciate the fix.

PR #276, opened 2024-03-03
Fixes #275

### LINKED ISSUE #275: Table sort icon does not appear when table column  "value" is used "name" is different from  "value"
This is a small graphical bug which shows up when trying to use both a "name" and a "value" parameter in the Table component. In this case, the arrow showing which direction is sorted simply does not show  up. 

For instance if you want to use capitalized column names in the \<Table\> "Order" example as below, the icon does not show up when clicking the column header.
```
columns={[
      {
        name: 'Name',
        value: 'name',
        align: 'left',
      },
     ...
]
```

jpoles1:
This is a quick fix that should rectify #275!

### DIFF
--- a/packages/svelte-ux/src/lib/components/TableOrderIcon.svelte
+++ b/packages/svelte-ux/src/lib/components/TableOrderIcon.svelte
@@ -9,7 +9,7 @@
   export let column: ColumnDef;
 </script>
 
-{#if order.by && (order.by === column.name || order.by === column.orderBy)}
+{#if order.by && (order.by === column.value || order.by === column.name || order.by === column.orderBy)}
   <div
     class="TableOrderIcon inline-block transition duration-100 transform"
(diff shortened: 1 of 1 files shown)


techniq (approved):
Thanks @jpoles1.  I kind of [forgot](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/types/table.ts#L20) `value` can be a `string` as well as a function (I always use it as a function).  When a string it can also be a "dotted" path (`foo.bar.baz`) (see [getCellValue()](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/utils/table.ts#L102-L108))

With that said, I usually use `name` and `header` to do similar, instead of `name` and `value`..

```
columns={[
      {
        name: 'name',
        header: 'Name',
        align: 'left',
      },
     ...
]
```

but worth fixing this way as well.  Thanks!

PR #277, opened 2024-03-03
Sorting + pagination example

jpoles1:
Adding a demonstration showing how you can use both pagination and sorting together; hopefully will be useful to someone else. The issue is that if simply combining the two functionalities together, sorting will only occur on the currently active page. If you use a separate reactive declaration and sort that however, the sorting goes across multiple pages.

Hopefully it's a useful addition to the documentation showing a common use case, but if not helpful/doesn't belong, that's OK too!

### DIFF
--- a/packages/svelte-ux/src/routes/docs/components/Table/+page.svelte
+++ b/packages/svelte-ux/src/routes/docs/components/Table/+page.svelte
@@ -30,8 +30,6 @@
     { id: 13, name: 'Oreo', calories: 437, fat: 18.0, carbs: 63, protein: 4.0 },
   ];
 
-  $: sorted_data = data.sort($order.handler);
-
   function randomDataGen() {
     return data.map((d) => {
       return {
@@ -206,7 +204,7 @@
 <h2>Order + Pagination</h2>
 
 <Preview>
-  <Paginate items={sorted_data} perPage={5} let:pageItems let:pagination>
+  <Paginate items={data.sort($order.handler)} perPage={5} let:pageItems let:pagination>
     <Table
       data={pageItems}
       columns={[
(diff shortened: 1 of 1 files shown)


packages/svelte-ux/src/routes/docs/components/Table/+page.svelte
```diff
@@ -201,6 +203,51 @@
   />
 </Preview>
 
+<h2>Order + Pagination</h2>
+
+<Preview>
+  <Paginate items={sorted_data} perPage={5} let:pageItems let:pagination>
```
techniq, line 209: It might make the example easier to see if the sorting was done in the markup, instead of a separate variable.

```svelte
<Paginate items={data.sort($order.handler)} perPage={5} let:pageItems let:pagination>
```

packages/svelte-ux/src/routes/docs/components/Table/+page.svelte
```diff
@@ -201,6 +203,51 @@
...
+        },
+        {
+          name: 'protein',
+          align: 'right',
+          format: 'integer',
+        },
+      ]}
+      orderBy={$order.by}
+      orderDirection={$order.direction}
+      on:headerClick={(e) => order.onHeaderClick(e.detail.column)}
```
techniq, line 240: When changing the sort order (via `onHeaderClick`), it usually makes sense to reset back to page one (if you were on say page 2 or 3).

Something like this should work:
```svelte
on:headerClick={(e) => {
  pagination.setPage(1);
  order.onHeaderClick(e.detail.column)
}}
```

techniq (changes requested):
I love adding more examples, thanks!  Suggested a few refinements, let me know what you think.

techniq (approved):
Looks good, thanks @jpoles1!

Issue #186, opened 2024-01-05
Handle svelte-check warnings


techniq:
- [x] Handle all `svelte-check` warnings (A11y, unused variables, etc) so the console is clean
  - Possibly leverage `<!-- svelte-ignore ... -->` **temporarily**
- [x] Setup CI to run `svelte-check` and fail build
- [ ] If `<!-- svelte-ignore ... -->` or `// @ts-expect-error` are used, properly handle remaining items
- [x] Rinse/repeat in [LayerChart](https://github.com/techniq/layerchart)
- [x] Enjoy a peaceful dev experience with improved accessibility

---

Reference
- `pnpm check --no-tsconfig --ignore "dist,src/routes`
- https://www.w3.org/WAI/ARIA/apg/patterns/
  - https://www.w3.org/WAI/ARIA/apg/example-index/
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA
- https://www.radix-ui.com/primitives/docs/overview/introduction
- https://react-spectrum.adobe.com/react-aria/Button.html
- https://melt-ui.com/docs/introduction

jycouet:
in user land, if you add in `svelte.config.js` the function `onwarn`: 

```js
/** @type {import('@sveltejs/kit').Config} */
const config = {
	// ...

	onwarn(warning, defaultHandler) {
		// Do not show 3rd party warnings
		if (warning.filename.includes('node_modules')) return

		defaultHandler(warning)
	},

	// ...
}

export default config
```

I find it nice until it's "fixed" 😉

techniq:
Making good progress in PR #409

techniq:
![zoolander-will-ferrell](https://github.com/techniq/svelte-ux/assets/177476/fe51a2fa-4556-4397-a3fe-5c8ccc2a358d)

![image](https://github.com/techniq/svelte-ux/assets/177476/0d84af15-9128-43c5-af60-534480cd9309)

![image](https://github.com/techniq/svelte-ux/assets/177476/91953b8d-a613-41e7-962d-bc260433ac98)

jycouet:
[...]

PR #281, opened 2024-03-06
Support NodeNext packages

### LINKED ISSUE #280: Support `NodeNext` packages
Projects and packages using `"moduleResolution": "NodeNext"` are unable to use `svelte-ux` or `layerchart` due to these projects not importing with the `.js` extension. The simplest way to resolve this is to set the projects `"moduleResolution": "NodeNext"` and fix the imports to prevent regression.

References:
	- https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution-for-libraries
	- https://kit.svelte.dev/docs/packaging#caveats

techniq:
Resolves #280

### DIFF
--- a/.changeset/cold-pianos-rest.md
+++ b/.changeset/cold-pianos-rest.md
@@ -1,0 +1,5 @@
+---
+"svelte-ux": minor
+---
+
+Support `NodeNext` packages
--- a/packages/svelte-ux/src/docs/Blockquote.svelte
+++ b/packages/svelte-ux/src/docs/Blockquote.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import { mdiInformation } from '@mdi/js';
   import Icon from '../lib/components/Icon.svelte';
-  import { cls } from '$lib/utils/styles';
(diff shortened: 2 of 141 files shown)


techniq:
* **#281** <a href="https://app.graphite.dev/github/pr/techniq/svelte-ux/281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @techniq and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

techniq:
`pnpm check | grep -B 1 -A 3 "Cannot find module"` still has 4 failures in the root [+page.svelte](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/routes/+page.svelte)...

image

image

but since it's only docs related and doesn't break the build (until we resolve all warnings/errors in the #186) I think it's OK to ignore for now (could have even been failing prior to the change).

PR #290, opened 2024-03-11
Support classes.icon root and path in NavItem

vanderburg:
Added support for classes.icon to be an object to pass to classes instead of class for the Icon. I don't love ternaries, but I didn't want the change to be breaking, so it still supports the original strings. Maybe there's a cleaner way to do this; could probably just make extra variables.

### DIFF
--- a/packages/svelte-ux/src/lib/components/NavItem.svelte
+++ b/packages/svelte-ux/src/lib/components/NavItem.svelte
@@ -17,7 +17,7 @@
     root?: string;
     active?: string;
     icon?: {
-      root?: string |
+      root?: string,
       path?: string | string[]
     };
   } = {};
--- a/packages/svelte-ux/src/lib/components/NavItem.svelte
+++ b/packages/svelte-ux/src/lib/components/NavItem.svelte
@@ -16,8 +16,8 @@
   export let classes: {
     root?: string;
     active?: string;
-    icon?: string | {
-      root?: string |;
+    icon?: {
+      root?: string |
       path?: string | string[]
     };
   } = {};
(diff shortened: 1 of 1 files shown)


packages/svelte-ux/src/lib/components/NavItem.svelte
```diff
@@ -16,7 +16,10 @@
   export let classes: {
     root?: string;
     active?: string;
-    icon?: string;
+    icon?: string | {
+      root?: string |;
```
techniq, line 20: Bad `;` character.  We can also change this to `icon?: ComponentProps<Icon>['classes']` (ex. [Card](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Card.svelte#L18)

I'm OK with making this a breaking change... I don't know if anyone would be using `<NavItem classes={{ icon: '' }}>` and I think breaking now instead of continuing to support the use case is best.

techniq (changes requested):
Thanks for the PR @vanderburg.  Small changes to simplify and we should be good.

techniq:
Thanks @vanderburg.  I merged and then made a few things.

- Ran prettier to fix the lint failures (https://github.com/techniq/svelte-ux/commit/26ad780503d9b266cc7f5afc19c7e0af44d65fe0).  My editor formats on save, but you can also run `pnpm format` to format the full repo, which will include any changes you've made (and should only update those files since the others should be a no-change
- Use `ComponentProps<Icon>['classes']` for `classes.icon` type (so they always match)
- Added a changeset using `pnpm changeset`.  This adds an line to the changelog and helps determine the next version to publish.  All PRs should include this to make sure a new version will be published, and we know what changed.

I appreciate the PR.  Going to publish now :)

PR #297, opened 2024-03-15
Implement type-safe `fromEntries` method

brandonmcconnell:


brandonmcconnell:
Fixed that bug in [6588220](https://github.com/techniq/svelte-ux/pull/297/commits/6588220989d44309671ab3da8b055124c754e39e). Should be good to go now.

techniq:
Thanks @brandonmcconnell.  Can you add a changeset `(pnpm changeset)` so it gets added to the changelog and updated the package version?

PR #301, opened 2024-03-22
Fix: Make ScrollFadeOptions optional in `scrollFade` action

brandonmcconnell:
What this fixes: TS error when using `use:scrollFade` without args:

Current workaround


techniq (approved):
Thanks!

PR #304, opened 2024-03-23
Add support for wildcard import paths

brandonmcconnell:
## ⚠️  Experimental

I am not sure if this works or not without testing this within another repo, but the general idea is fixing the issue where the def and type exports are seen as coming from different import paths (w/ vs. w/o `dist`):

```ts
import { sumObjects as _sumObjects } from 'svelte-ux/utils/array';
import type { sumObjects as SumObjects } from 'svelte-ux/dist/utils/array';

const sumObjects = _sumObjects as typeof SumObjects;
```

Whenever you find time to test this, I think this might help to resolve that issue. More of a theory based on the docs here:
https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports

### DIFF
--- a/packages/svelte-ux/package.json
+++ b/packages/svelte-ux/package.json
@@ -78,11 +78,13 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "svelte": "./dist/index.js"
+      "svelte": "./dist/index.js",
+      "default": "./dist/*"
     },
     "./utils/*": {
       "types": "./dist/utils/*.d.ts",
-      "svelte": "./dist/utils/*.js"
+      "svelte": "./dist/utils/*.js",
+      "default": "./dist/utils/*"
     },
     "./plugins/*": "./dist/plugins/*",
     "./styles/*": "./dist/styles/*"


techniq:
hmm... I haven't seen this needed in other popular Svelte packages ([superforms](https://github.com/ciscoheat/sveltekit-superforms/blob/main/package.json#L49-L70), [melt-ui](https://github.com/melt-ui/melt-ui/blob/develop/package.json#L29-L55), [bits-ui](https://github.com/huntabyte/bits-ui/blob/main/package.json#L24-L29), ...).

I have seen VSCode try to import from `/dist` but it should always be removed.  I've used `import { type ... } from 'svelte-ux'; in other packages without an issue.

Not exactly sure what you are trying to do here...
```
import { sumObjects as _sumObjects } from https://github.com/techniq/svelte-ux;
import type { sumObjects as SumObjects } from https://github.com/techniq/svelte-ux;

const sumObjects = _sumObjects as typeof SumObjects;
```

brandonmcconnell:
_Responded in video chat, @techniq was able to reproduce the issue_

techniq:
Going to close for now as not sure we need this (seems to only be a typescript issue with Svelte UX utils within non-Svelte projects).  We can investigate more, but don't have much time ATM.

PR #305, opened 2024-03-25
Allow setting Menu transition configuration from `settings`

dimfeld:
Would probably be useful to add this for other components as well but that may require some additional prop drilldown I think. This is a good start to just customize everything that uses Menu, which I think will be the most common case.


techniq (approved):
Thanks @dimfeld, this is something I've wanted to add.

Setting more of the default values in reactive statements does confuse what the default values are (this one was already confusing TBH), but I think embracing jsdocs comments (which will help Sveld) will help this situation.

Issue #250, opened 2024-02-12
Design token follow-up


techniq:
- [ ] Support passing color objects from tailwind
- [ ] Verify `create-svelte-ux` / templates
- [ ] Allow all colors to be optional (there exists a default light/dark theme if none present)
- [ ] Set accent: hsl(var(--color-primary)) by default?
- [ ] Change default state colors to tailwind colors, still set daisy UI specific state colors in convertor
- [ ] Migration guide (`accent` color to `primary`)

brandonmcconnell:
"Support passing color objects from tailwind" will be solved by #307

PR #307, opened 2024-03-25
Add support for color objects with nested keys to `processThemeColors` via new `flattenThemeColors` function

brandonmcconnell:
This will resolve "Support passing color objects from tailwind" from #250


packages/svelte-ux/src/lib/styles/theme.ts
```diff
@@ -65,43 +65,50 @@ export function getThemeNames(themes: Record<string, any>) {
...
+  keys: (string | number)[] = [],
+  memo?: Record<string, string>
 ) {
-  const colors = { ...themeColors };
+  return entries(themeColors).reduce<Record<string, string>>((memo, [key, value]) => {
+    const newKeys = [...keys, key];
+    if (typeof value === 'string') {
+      memo[newKeys.join('-')] = value;
+    } else {
+      flattenColors(value, newKeys, memo);
```
techniq, line 85: Is there a need for this to be recursive (can you provide an example)?

If not, I would like to simplify this to something like...
```js
return fromEntries(entries(colors).flatMap(([color, value]) => {
	if (typeof value === 'string') {
		return [color, value];
	} else {
		return entries(value).map(([shade, value]) => {
			return [`${color}-${shade}`, value]
		})
	}
}))
```

brandonmcconnell: I use a similar optimization in my active PR:


I added support for recursion so if anyone wants to pass in an object that contains multiple levels, they don't have to do the flattening themselves, e.g.:
```ts
const colors = {
  gray: {
    cool: {
      50: '#f8fafc',
      100: '#f1f5f9',
      200: '#e2e8f0',
      300: '#cbd5e1',
      400: '#94a3b8',
      500: '#64748b',
      600: '#475569',
      700: '#334155',
      800: '#1e293b',
      900: '#0f172a',
      950: '#020617',
    },
    warm: {
      50: '#fafaf9',
      100: '#f5f5f4',
      200: '#e7e5e4',
      300: '#d6d3d1',
      400: '#a8a29e',
      500: '#78716c',
      600: '#57534e',
      700: '#44403c',
      800: '#292524',
      900: '#1c1917',
      950: '#0c0a09',
    },
  }
}
```

This same idea is also actively being discussed for inclusion in CSS (primarily for use in CSS libs like Tailwind CSS) here:
https://github.com/w3c/csswg-drafts/issues/9992

brandonmcconnell: @techniq If you would still rather this only support one level, I can refactor this. Just lmk

techniq: I'm fine with the recursion.  A lot of the times I think of `-content` as it's own color and reference it explicitly (`primary-content`, etc) but supporting it in the same object is nice.

Care to add a test for `flattenColors()` (create `theme.test.ts` ) to capture these use cases?  I need to add tests for more of the theme functions (`processThemeColors()`, etc), but this would be a good and simple start.  Thanks!

brandonmcconnell: Tests added for both `flattenColors` and `processThemeColors`, and support added for `'DEFAULT'` shade, matching Tailwind's nomenclature, though this will be usable by any CSS library.

techniq (approved):
Looks good.  Thanks!

PR #316, opened 2024-03-30
Docs: Switch to enable realtime theme application

dimfeld:
CleanShot 2024-03-30 at 13 04 59

Will probably have something to store the custom theme in `sessionStore` coming up soon.

### DIFF
--- a/packages/svelte-ux/src/routes/theme/+page.svelte
+++ b/packages/svelte-ux/src/routes/theme/+page.svelte
@@ -4,6 +4,7 @@
   import Button from '$lib/components/Button.svelte';
   import ButtonGroup from '$lib/components/ButtonGroup.svelte';
   import CopyButton from '$lib/components/CopyButton.svelte';
+  import Field from '$lib/components/Field.svelte';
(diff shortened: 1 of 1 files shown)


dimfeld:
Not totally happy with the label here, feel free to change it if you have a better idea

techniq:
Thanks @dimfeld! 

I think at some point my plan was to register `custom-light` and `custom-dark` themes instead of overriding `light` / `dark` (like it did before as well) in case you wanted to restore the original daisy themes.  ATM you can just refresh to get them back, but once we allow persisting these values, I think this will be best.  We can probably just add these themes to the [list](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/routes/%2Blayout.server.ts#L17) (but it's been a bit since I've looked deeply at this :) ).  Also, I'll update the label :) (maybe just "Enable preview" like Skeleton uses).  I think we can remove the explicit "Override current themes" button to (can't think of any reason to click it manually now).

Going to merge this as it's already an improvement

PR #317, opened 2024-03-30
Remove unneeded postcss-import load

dimfeld:
The Neovim Svelte plugin was failing since it couldn't load postcss-import, referenced from the postcss config. 

Not sure why there weren't issues anywhere else, but it doesn't seem like anything is actually doing an `@import` and everything appears to build fine without it.

### DIFF
--- a/packages/svelte-ux/postcss.config.cjs
+++ b/packages/svelte-ux/postcss.config.cjs
@@ -1,6 +1,5 @@
 module.exports = {
   plugins: {
-    'postcss-import': {},
     'tailwindcss/nesting': {},
     tailwindcss: {},
     autoprefixer: {},


techniq:
Yeah, I've never used this directly, although 9 months ago I [added](https://github.com/techniq/svelte-ux/commit/7b9a2e10a1243b30fa93910c379c8c28fdd73f5e) it to "Fix tailwind warning about nesting plugin".  I see it in their [docs](https://tailwindcss.com/docs/using-with-preprocessors#build-time-imports), but only if using it directly from what can tell.  I'll remove it but keep an eye out if something fails unexpectedly.   If so, I guess we'll just need to add the `postcss-import` package.

I did just notice it is a tailwind dependency (saw it in [pnpm-lock.yaml](https://github.com/techniq/svelte-ux/blob/97d20586512af2001bc899b9061a3fe20322c1c7/pnpm-lock.yaml#L4123)) but once again, Svelte UX doesn't use it directly.

PR #320, opened 2024-03-31
Docs: Export buttton only writes the explicit settings, and converts to selected color space

### LINKED ISSUE #318: Some way to copy themes to clipboard with only the explicitly set variables.
If you copy a theme to the clipboard, it copies all the variables, including all the ones that the plugin calculates for you. I think it would be potentially more useful to only copy the explicitly configured values out.

The primary benefit is that it becomes easier to tweak the generated theme posthoc and have the plugin then propagate the changes to all the other shades. Right now you can do that by deleting a bunch of the generated theme variables from the tailwind config but it's not clear right away which ones should be removed.

dimfeld:
Resolves #318 

I also added a "Copy Full Theme" menu item which retains the old behavior. Seems like it's kind of inconsistent though... when you start from a Skeleton theme then you seem to get the full exported theme, but when you start from a Daisy theme then the export is much shorter. I figure you will have a good idea of why that is :) Or maybe the skeleton themes are just including a whole bunch of variables that aren't used? Not totally sure. We may want to just remove this button.

This also changes the list of ColorFields to be data-driven since I was already creating an array of theme keys anyway as part of this.

### DIFF
--- a/packages/svelte-ux/src/routes/theme/+page.svelte
+++ b/packages/svelte-ux/src/routes/theme/+page.svelte
@@ -24,6 +24,7 @@
   } from '$lib/styles/theme.js';
   import { entries, fromEntries, type MenuOption } from '$lib/types/index.js';
   import ColorField from './ColorField.svelte';
+  import { formatColor } from './colors.js';
   import { getSettings } from '$lib/components/settings.js';
 
   export let data;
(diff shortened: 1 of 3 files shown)


techniq:
Actually I just looked in my Strava / Github analysis projects and we don't have to be explicit for `dark` anymore (just for additional dark themes) so we should be good.

dimfeld:
Added `color-scheme` back

techniq:
Thanks @dimfeld.  I'll probably remove "Copy Full Theme" in the future once border radius theming is implemented as I can't think of a good reason to keep it (and reduce confusion)

PR #321, opened 2024-03-31
Docs: Persist custom theme

dimfeld:
This is stacked on top of #320 so currently includes changes from both.

This still needs a little more work, or we may want to drop this in favor of a better solution that addresses these pain points:

- The theme is saved, but when you reload the page custom themes are not applied unless you go back to the theme page and reenable preview. I think if we want to re-apply on startup it may require some higher-level integration of the custom theme.
- It now always says custom instead of the actual theme name... probably need to persist the selected theme value as well or something so we can reload that when it's not custom.


techniq:

PR #324, opened 2024-04-05
Update `sumObjects` to coerce numeric strings, preserve Arrays in return

brandonmcconnell:
Changed made
- Update `sumObjects` to automatically coerce numeric strings to numbers
- Update `sumObjects` to return the summed values as an Array if all processed items are arrays

### DIFF
--- a/packages/svelte-ux/src/lib/utils/array.ts
+++ b/packages/svelte-ux/src/lib/utils/array.ts
@@ -43,10 +43,11 @@ export function sumObjects(items: (object | null)[], prop?: PropAccessorArg) {
 
   const result = rollup(
     items.flatMap((x) => entries(x ?? {})),
-    (values) => sum(values, (d) => {
-      const value = Number(getProp(d[1]));
-      return Number.isFinite(value) ? value : 0;
-    }),
+    (values) =>
+      sum(values, (d) => {
+        const value = Number(getProp(d[1]));
+        return Number.isFinite(value) ? value : 0;
+      }),
     (d) => d[0]
   );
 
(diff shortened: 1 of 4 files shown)


packages/svelte-ux/src/lib/utils/array.ts
```diff
@@ -38,14 +38,19 @@ export function sum(items: (object | null)[], prop?: PropAccessorArg) {
...
-      (values) => sum(values, (d) => (Number.isFinite(d[1]) ? d[1] : 0)),
-      (d) => d[0]
-    )
+export function sumObjects(items: (object | null)[], prop?: PropAccessorArg) {
+  const getProp = propAccessor(prop);
+
+  const result = rollup(
+    items.flatMap((x) => entries(x ?? {})),
+    (values) => sum(values, (d) => {
+      const value = Number(getProp(d[1]));
```
techniq, line 47: I wonder if this should also get the `key` as the second argument (`d[0]` and maybe the whole object as the third (`d`), similar to [just-map-values](https://github.com/angus-c/just?tab=readme-ov-file#just-map-values)

techniq (changes requested):
Thanks @brandonmcconnell.  Could you run `pnpm format` to fix the lint issues (prettier formatting) which will allow the CI tests to run, and add a few more tests...

- Using the accessor as a function (ex. `sumObjects(data, d => ...)`)
- Using the accessor as a string (ex. `sumObjects(data, 'value')`)
  - data: `[{ foo: { value: 123 }, bar: { value: 234 } }, { foo: { value: 987 }, ... }]`
- Data as an array of arrays

Thanks!

techniq (approved):

PR #331, opened 2024-04-21
Add missing settingsClasses to Table.svelte

29th-Day:
Table isnt stylable since the settingsClass isnt loaded. I noticed it in my own project, I dont know is its all changes needed.

### DIFF
--- a/packages/svelte-ux/src/lib/components/Table.svelte
+++ b/packages/svelte-ux/src/lib/components/Table.svelte
@@ -12,6 +12,8 @@
   import { getSettings } from './settings.js';
   import tableOrderStore from '../stores/tableOrderStore.js';
 
+  const settingsClasses = getComponentClasses('Table');
+
   type T = $$Generic;
 
   const dispatch = createEventDispatcher<{


techniq:
Thanks @29th-Day, but there is actually a commented out `const settingsClasses = getComponentClasses('Table');` already in `Table.svelte` due to a circular reference I haven't been able to dig into and solve yet.

image

If you could find the solution I would love to enable this.  Also, this PR has `settingsClasses` declared twice

image

techniq:
@29th-Day btw, I'm pretty sure the issue is using `settingsClasses` within the reactive variables `headers` and `rowColumns`.  Not sure the best solution ATM, but it's solvable.

https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Table.svelte#L52-L71

techniq:
also not sure why this creates a circular reference either.

techniq:
Tracking via https://github.com/techniq/svelte-ux/issues/332 for now

Issue #29, opened 2022-11-04
Popover improvements (including Menu, Tooltip)


techniq:
- [x] Portal support
  - Fix z-index issues (see Tooltip [issue](https://github.com/techniq/svelte-ux/issues/29), but also Menu usage will be fixed)
- [x] Flip (opposite side if scrolled outside of view)
  - https://floating-ui.com/docs/flip
- [x] Auto placement (most available space)
  - Based on if there is more room above or below the page (if not explicitly specified)
  - https://floating-ui.com/docs/autoplacement
- [ ] `maxViewport` / `size`
  - https://floating-ui.com/docs/size
- [ ] Arrow
  - https://floating-ui.com/docs/arrow
  - https://css-tricks.com/snippets/css/css-triangle/
  - [Simple tooltips with no extra pseudo-element](https://codepen.io/thebabydino/pen/dyKZRzy)
  - [Fancy & Animated Tooltip (CSS Only)](https://codepen.io/reiinii1/pen/aPGXEa)
- [ ] Hide if offscreen
  - https://floating-ui.com/docs/hide

myieye:
@techniq I'm trying to use (and am very much enjoying 😃) svelte-ux in a web-component. I'm embedding all the styles in the web-component shadow DOM. As a result, the popovers that are added as children of `document.body` don't get the styles they need. The portal is exposed on some components (e.g. `Drawer`), but not on all (e.g. `MultiSelectField`).

What I'd **really love** is a way to change the [default/fallback portal target](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/actions/portal.ts#L45). I'd be happy to try to contribute the change, but I'm having trouble figuring out what the right way to do it would be.

I first thought I'd add it as a [setting](https://github.com/techniq/svelte-ux/blob/513dcc29605955a01641cd255920a780dbe79eb4/packages/svelte-ux/src/lib/components/settings.ts), but I don't think it's really allowed/works to use the context API in an action. Do you have a suggestion?

techniq:
Hey @myieye 👋, thanks for checking out Svelte UX, and interesting use case (I hadn't heard of anyone using it within a web component, but I'm very intrigued).

Yeah, accessing context within an action is not supported (afaik as well).  My initial thought is you really just need to define/override the target for component's that use the `portal` action internally (like `Drawer`, `Dialog`, etc).

[...]

PR #333, opened 2024-05-02
Add default portal target setting

myieye:
Seems to be working. I need this change in order to use svelte-ux in a shadow DOM/web component

image


myieye:
[Decided on a different implementation](https://github.com/techniq/svelte-ux/issues/29#issuecomment-2091476032).

PR #334, opened 2024-05-03
Add well known PortalTarget class and examples

myieye:
Adds `PortalTarget` ([as discussed](https://github.com/techniq/svelte-ux/issues/29#issuecomment-2091476032)) as a well known class that can be used to globally customize the default portal target used by the portal action.

I also:
- Added examples of the different target approaches (see screenshot below)
- Changed the portal logic to put the portal back in its original parent if it becomes disabled (rather than doing nothing)
I expect this was just an oversight so far. If a portal is initially disabled, it just stays in its original parent, so it seems to make sense to consider that its "disabled state". And if a user doesn't want the portal to move, they just shouldn't change the options.
I primarily made the change, because it meant that I needed less code for the examples, but I think that it's the intuitive behaviour.

image

### DIFF
--- a/packages/svelte-ux/src/lib/actions/portal.ts
+++ b/packages/svelte-ux/src/lib/actions/portal.ts
@@ -7,18 +7,29 @@ export type PortalOptions =
     }
   | boolean;
 
+type PortalTargets = {
+  fallbackTarget: Element | null;
+  originalParent: HTMLElement | null;
+};
+
 /**
  * Render component outside current DOM hierarchy
  */
 export const portal: Action<HTMLElement, PortalOptions> = (node, options) => {
-  moveNode(node, options);
+  const targets = {
+    // prefer ancestors, but it doesn't have to be an ancestor
+    fallbackTarget: node.closest('.PortalTarget') ?? document.querySelector('.PortalTarget'),
+    originalParent: node.parentElement,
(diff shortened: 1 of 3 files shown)


packages/svelte-ux/src/lib/actions/portal.ts
```diff
@@ -7,18 +7,29 @@ export type PortalOptions =
...
+    // prefer ancestors, but it doesn't have to be an ancestor
+    fallbackTarget: node.closest('.PortalTarget') ?? document.querySelector('.PortalTarget'),
+    originalParent: node.parentElement,
+  };
+  moveNode(node, options, targets);
 
   return {
     update(options) {
-      moveNode(node, options);
+      console.log(options);
```
techniq, line 28: remove please :)

techniq (changes requested):
This looks great and I appreciate you adding examples to the docs, and I agree disabling should restore back to the original location, good catch.  Let a few little nits then looks to merge.

packages/svelte-ux/src/lib/actions/portal.ts
myieye: Woops. I removed it.

techniq:
Thanks @myieye!

PR #346, opened 2024-05-10
CopyButton enhancements

dimfeld:
- Allow `value` to be a function that returns a string
- "Copied!" message customization 
- Emit a `click` event when the button is clicked

### DIFF
--- a/packages/svelte-ux/src/lib/components/CopyButton.svelte
+++ b/packages/svelte-ux/src/lib/components/CopyButton.svelte
@@ -10,7 +10,7 @@
 
   const { classes: settingsClasses, defaults } = getComponentSettings('CopyButton');
 
-  export let value: string;
+  export let value: string | (() => string);
 
   let showMessage = false;
   $: if (showMessage) {
@@ -25,7 +25,8 @@
   {...restProps}
   class={cls('CopyButton', settingsClasses.root, className)}
   on:click={() => {
-    navigator.clipboard.writeText(value);
+    let copyValue = typeof value === 'function' ? value() : value;
+    navigator.clipboard.writeText(copyValue);
     showMessage = true;
   }}
 >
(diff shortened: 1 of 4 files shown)


packages/svelte-ux/src/lib/components/CopyButton.svelte
```diff
@@ -25,11 +29,15 @@
   {...restProps}
   class={cls('CopyButton', settingsClasses.root, className)}
   on:click={() => {
-    navigator.clipboard.writeText(value);
-    showMessage = true;
+    let copyValue = typeof value === 'function' ? value() : value;
+    navigator.clipboard.writeText(copyValue);
+    dispatch('click');
```
techniq, line 34: Instead of using `dispatch` maybe just add another `on:click` to forward the event.  It's a pattern I've used elsewhere ([NavItem](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/NavItem.svelte#L42), [ToggleButton](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/ToggleButton.svelte#L21))

techniq (commented):
Nice enhancements.  Suggestion to simplify forwarding the `on:click` event

packages/svelte-ux/src/lib/components/CopyButton.svelte
dimfeld: Yeah I go back and forth on that. Ended up doing it this way to be more explicit about the event firing after the clipboard has been copied, instead of relying on the undocumented behavior that the event handlers fire in the order they are listed in the template. (Not that it's likely to change at this point)

Also this is very slightly closer to how it will be in Svelte 5. :) Anyway I can change it if you still would like me to.

techniq: Nope, makes sense to me to have the explicit ordering, especially to align with how things will be going forward with Svelte 5, which I believe becomes...

```svelte
<script>
  const { onclick } = $props;
</script>

<Button
  onclick={() => {
    // ...
    onclick();
  }}
/>
```

(or something like that.  just wrote that in a github comment without checking too much).

techniq (approved):
Thanks @dimfeld

Issue #54, opened 2023-05-02
Add RangeField


techniq:
Currently there is RangeSlider

- [ ] Rename [RangeSlider](https://svelte-ux.vercel.app/docs/components/RangeSlider) to `RangeInput` (or maybe `MultiRangeInput`)
- [ ] Add `RangeField` which combines `Field` and `RangeInput`
- [ ] Update `RangeInput` to support single value
- [ ] Consider adding `native` prop which uses a styled `<input type="range" {min} {max} {step} />`
  - [Styled range input](https://svelte.dev/repl/cfa047ad97cf49c6bfb5b7e5881d5600?version=3.44.2)
- [ ] Consider consolidating `NumberStepper`
- [ ] Add ticks support
  - [Input range with labeled ticks](https://svelte.dev/repl/9e8b2964cfbc4751b0096d9d262cec44?version=3.58.0)
- [ ] Add `ColorRangeInput` and expose maybe as `type="color"`
  - [REPL](https://svelte.dev/repl/1bb1dac377594b52b5139abb6da37b84?version=3.58.0)
- [ ] Replace LayerChart's [RangeField](https://github.com/techniq/layerchart/blob/master/src/lib/docs/RangeField.svelte) (but leverage some of the ideas like showing the number, etc
  - Support clicking number to set (with proper mobile keyboard)
- [ ] Consider adding a [CircularSlider](https://svelte.dev/repl/e5db52ead53643c381eb626c7ee0f5a8?version=3.58.0) (or `CircleRangeInput`)

## Examples:
- https://observablehq.com/@mootari/range-slider#doc_interval
- https://www.radix-ui.com/docs/primitives/components/slider
- https://mhkeller.com/svelte-double-range-slider/
  - https://github.com/mhkeller/svelte-double-range-slider
- https://observablehq.com/@bumbeishvili/data-driven-range-sliders
- [Multi-range conjoined tooltip](https://codesandbox.io/embed/interesting-snyder-ttizkx)

PR #348, opened 2024-05-16
RangeSlider: add disableTooltips option

ddol:
# What changed?

Add option to `RangeSlider` component to disable tooltips. 

```jsx
  <RangeSlider disableTooltips />
```

Update documentation to provide example:
Screenshot 2024-05-15 at 7 21 32 PM

# How was it tested?

Development server was spun up on branch, changes were validated using:

```bash
pnpm i
cd packages/svelte-ux; pnpm dev
```

Unit tests were run but 6 tests failed. These tests also failed on `main`, so I'm assuming they are unrelated to this change.

### DIFF
--- a/.changeset/angry-months-matter.md
+++ b/.changeset/angry-months-matter.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+[RangeSlider] Add disableTooltips property
--- a/packages/svelte-ux/src/lib/components/RangeSlider.svelte
+++ b/packages/svelte-ux/src/lib/components/RangeSlider.svelte
@@ -41,6 +41,7 @@
   export let step = 1;
   export let value = [min, max];
   export let disabled = false;
+  export let disableTooltips = false;
 
   let className: string | undefined = undefined;
   export { className as class };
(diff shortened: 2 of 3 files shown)


techniq:
Works for me, although I wonder if we shouldn't default to not showing tooltips and opting in with `showTooltips` or similar.

ddol:
I think the tooltips are a good default as they give more clarity as to the values for the range, which is why I opted for a flag to disable. But I'm happy to change to `showTooltips` if you prefer.

techniq:
I'm fine with leaving it enabled by default for now.  I took a very quick look at a few other popular component libraries and it seems only MUI and Ant have tooltip support, MUI defaults to off (opt in with `valueLabelDisplay="auto"`) and Ant defaults to on (opt out with `tooltip={{ open: false }}`

- https://mui.com/material-ui/react-slider/#range-slider
- https://ant.design/components/slider
- https://ui.shadcn.com/docs/components/slider
- https://www.radix-ui.com/primitives/docs/components/slider

I'd like to revisit `RangeInput` / `RangeField` / etc (https://github.com/techniq/svelte-ux/issues/54) at some point and make it both simplifier and more flexible, but no need to tackle that now.  

Thanks for the PR!

techniq:
Oh, could you add a changeset (`pnpm changeset`)

PR #353, opened 2024-05-18
Fix create-svelte-ux on windows

29th-Day:
When using "npm create svelte-ux@latest" on a windows machine it wouldn't work. Not tested on Linux thou.

### DIFF
--- a/.changeset/wet-experts-wait.md
+++ b/.changeset/wet-experts-wait.md
@@ -1,0 +1,5 @@
+---
+'create-svelte-ux': minor
+---
+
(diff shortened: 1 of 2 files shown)


techniq:
Pulled locally and tested (Mac).

image

Thanks @29th-Day, I REALLY appreciate you fixing this!  The build is breaking due to prettier/lint and pnpm version on Vercel.  I fixed the later on `main` just now, and can fix the prettier issue after I merge (just running `pnpm format` should fix it).

PR #358, opened 2024-05-22
Fix overflow in starter template nav slots

ttytm:
Fixes an overflow resulting in a scrollbar. 

E.g., current after creating the starter template (changed the nav bg to red to increase contrast):

https://github.com/techniq/svelte-ux/assets/34311583/70df67aa-004e-4778-b0f9-081d6932fb08

With the changes, the same property class (`py-2`) is used as on the libs website, which won't result in an overflow as long as the nav items fit into the container.

---

_Related to this, having looked into the AppLayouts slots for the navigation_: would it be too pedantic to change the nav slot in the templates to remove the redundant use of `nav` tags in a follow-up PR?

E.g., current:
Screenshot_20240522_230535

https://github.com/techniq/svelte-ux/blob/e227b188736b7c4bdfa62e27f80f646f67fbf955/packages/create-svelte-ux/templates/starter/src/routes/+layout.svelte#L27

Afaik the following could work without having to make compromises:

```diff
-<nav slot="nav" class="nav h-full">
+<div slot="nav" class="h-full">
```

### DIFF
--- a/packages/create-svelte-ux/templates/layerchart/src/routes/+layout.svelte
+++ b/packages/create-svelte-ux/templates/layerchart/src/routes/+layout.svelte
@@ -8,7 +8,7 @@
 		classes: {
 			AppBar: 'bg-primary text-white shadow-md',
 			AppLayout: {
-				nav: 'bg-neutral-800'
+				nav: 'bg-neutral-800 py-2'
 			},
 			NavItem: {
 				root: 'text-sm text-gray-400 pl-6 py-2 hover:text-white hover:bg-gray-300/10 [&:where(.is-active)]:text-sky-400 [&:where(.is-active)]:bg-gray-500/10'
@@ -24,7 +24,6 @@
 			text="Home"
 			icon="M10,20V14H14V20H19V12H22L12,3L2,12H5V20H10Z"
 			currentUrl={$page.url}
-			class="mt-2"
 		/>
 
 		<NavItem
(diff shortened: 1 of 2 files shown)


techniq:
Thanks, I haven't looked at the templates closely since they were created, so aligning with the docs to fix overflow sounds great!  Thanks for looking into this.

Also, removing the nested `nav` and having semantic markup sounds like a good win (as long as it doesn't cause any regressions, which I don't anticipate any).

While looking into `AppLayout`, one thing I've been meaning to dig into is fixing anchor tags not scrolling to the item on direct routing, but works when navigated while on the page.  For example: https://svelte-ux.techniq.dev/docs/components/Button#text-with-icon does not jump down the page if you go to it directly, but clicking on the "Text with Icon" link does.

I think this stems from the [fact](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/AppLayout.svelte#L86) that `<main>` has `overflow: auto` and it's not just the `<body>` scrolling.  IIRC coordinating the overflow with `nav` and sticky content caused some difficulties, and just never dedicated the time.  If you have and willing to take a fresh look at it, I would appreciate it.

Thanks again for this PR.  I'm going to merge now and we can follow up in other PRs / issues.

techniq (approved):

ttytm:
Thank you! I just stumbled over the library - incredible work 🙇‍♂️

Gonna keep my eyes open when using it. It's great already!

PR #359, opened 2024-05-23
Update nav slot tags in `AppLayout`s to improve markup semantics

ttytm:
I think using a `svelte:fragment` tag here could be the cleanest solution for the default, resulting in the least amount of redundancies in required class assignments and tags in the final markup.

Refs.: https://svelte.dev/docs/special-elements#svelte-fragment, https://learn.svelte.dev/tutorial/svelte-fragment

If the suggestion is appropriate, please also update the PR message if there is something that fits better.

### DIFF
--- a/packages/create-svelte-ux/templates/layerchart/src/routes/+layout.svelte
+++ b/packages/create-svelte-ux/templates/layerchart/src/routes/+layout.svelte
@@ -18,7 +18,7 @@
 </script>
 
 <AppLayout>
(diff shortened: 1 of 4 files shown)


techniq:
Great suggestion, thanks @ttytm!  I also updated my other usage / references (LayerChart, Github Analysis, and Strava Analysis).

techniq (approved):

PR #360, opened 2024-05-25
Simplify copy in `create-svelte-ux/bin.js` and fix `.gitignore` not being copied

ttytm:
Simplifies the recursive `copy` function that is used in `npm create svelte-ux`.

The changes also fix not copying the `.gitignore` file. While the current code intends to copy the `.gitignore` file, it is missing in the result. This makes it bothersome to recreate the correct ignore file manually when initializing a git project after using the create script.

### DIFF
--- a/packages/create-svelte-ux/bin.js
+++ b/packages/create-svelte-ux/bin.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+
 import * as p from '@clack/prompts';
 import { program, Option, InvalidArgumentError } from 'commander';
 import { bold, cyan, gray, green, italic } from '@kitql/helpers';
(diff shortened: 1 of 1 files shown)


packages/create-svelte-ux/bin.js
```diff
@@ -120,8 +121,7 @@ copy(
     PROJECT_NAME: projectName,
     SVELTE_UX_VERSION: version,
   },
-  { '.meta.gitignore': '.gitignore' },
```
ttytm, line 123: If that helps with the review: 

For local comparison, to just fix copying the ignore file:

```diff
- { '.meta.gitignore': '.gitignore' },
+ { '.meta..gitignore': '.gitignore' },
```

Then the results of the updated script can be compared with the old fixed script and verified.

techniq (approved):
LGTM, thanks!

ttytm:
Testing the updated published script.
It looks like it behaves different in the published create package. Likely that's the reason why there is the `meta..gitignore` in the first place. I'd restore the object with the fix then.

PR #361, opened 2024-05-25
forward props from DatePickerField to inner components

abdullahcalisir12:


techniq (changes requested):
Could you remove the package-lock file and add a patch level changeset using 'pnpm changeset' to allow the NPM version to be bumped when merged?  Thanks!

techniq:
Thanks for the PR, but after looking a little closer at it, I went a slightly different approach.  There is now a standard `disabledDates` prop used by all the Date fields (also includes types since you can specify a function, array of dates, or start/end object)

see: https://github.com/techniq/svelte-ux/commit/c04a5aa286eb82b55225f61f1d15b57a180e4113

PR #369, opened 2024-05-26
Improve search accessibility and make minor style updates

ttytm:
The changes resolve some TODOs for the search field and improve accessibility.

The updates regard:
- Persistent/sticky search input field
- Keyboard navigation fixes
- Closing the search with esc
- Alignment in small viewports

---

I made a short screencast, hoping it will help as a quick and convenient way to present go over the changes.

https://github.com/techniq/svelte-ux/assets/34311583/4768e47d-2a4a-4cdf-9deb-241f48e53884

### DIFF
--- a/.changeset/late-carrots-draw.md
+++ b/.changeset/late-carrots-draw.md
@@ -1,0 +1,5 @@
+---
+"svelte-ux": minor
+---
+
+Improve search accessibility and make minor style updates
--- a/packages/svelte-ux/src/lib/components/QuickSearch.svelte
+++ b/packages/svelte-ux/src/lib/components/QuickSearch.svelte
@@ -66,21 +66,21 @@
     backdrop: 'backdrop-blur-sm',
   }}
 >
-  <div class="overflow-auto max-h-[min(90dvh,600px)] w-[400px] max-w-[95vw] py-1">
-    <SelectField
-      icon={mdiMagnify}
-      placeholder="Search..."
-      inlineOptions={true}
-      {options}
-      {fieldActions}
-      on:change
-      on:change={() => (open = false)}
-      classes={{
-        field: {
-          container: 'border-none hover:shadow-none group-focus-within:shadow-none',
-        },
-        group: 'capitalize',
-      }}
-    />
-  </div>
+  <SelectField
+    icon={mdiMagnify}
+    placeholder="Search..."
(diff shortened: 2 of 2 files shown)


techniq:
This is a great improvement, thanks @ttytm. 

I do wonder if the dialog should be taller, particularly on larger viewport, but even iPhone size feels like it could use a little more height...

image

When I played around with responsive, it didn't feel like the min of `90dvh` was kicking in (`max-h-[min(90dvh,380px)]`)

image

I mostly mention this if we decide to make the results/options taller than `380px` as this might cause an issue on smaller viewports not resizing below (but I could be overlooking something).

I'm going to go ahead and merge this as it is much better than it is now.  Thanks again!

ttytm:
Thank you for the review work!

Yeah the height is debatable. The driving factor was limiting the number of simultaneously presented result to make it easier to process for the viewers brain.

If there would be a description or other elements that give the single results more height I would have chosen a larger height already.
E.g., as a reference for other searches and the number of search results that provide at the same time the quick-search at the following sites: https://ui.shadcn.com, https://ui.shadcn.com, https://tailwindcss.com or just doing a regular bing / goolge search. 

Ultimately, I think it will remain a matter of preference, which is hard to argue, and I don't have a strong preference here either. So please update it to whatever fits best!

techniq:
@ttytm thanks for the explanation, I agree 😁.  I would like to pull all the meta data from the `+page.ts` files (including descriptions) for quick search at some point.  I know some sites use Algolia to index the context, but figure it's possible to walk/parse each of these files.

I do see some of the implementations center the search dialogs (shadcn, SvelteKit), but then the search input can jump for no results (shadcn), although SvelteKit's does not jump.  I think having it near the top is best, especially without having to solve the jumping issue (put it in a container) and is where it would be located on mobile anyways, just an observation.

I think what we have is great, and hopefully improve the display with descriptions and maybe some icons to better visual indications of each time (I like to think of icons as bullet points) over time.

Here's a few screenshots for later reference:

image

image

image

image

Thanks again!

ttytm:
Nice observations. I also liked the version close to the top better than the centered-"jumping" variant in shadcn (had it in there for a while before submitting the PR 🤓). Also agreed that it's great already and we can continue improving things step by step.

PR #371, opened 2024-05-29
Disable tooltips on print

brandonmcconnell:


techniq:
Curious how these get displayed in a print media query?

brandonmcconnell:
@techniq This brings print rendering closer to actual print by disabling tooltips when emulating print. The same clung be achieved with a `print:hidden` classname.

If you'd rather not add this, that's fine.

techniq:
> @techniq This brings print rendering closer to actual print by disabling tooltips when emulating print. The same can be achieved with a `print:hidden` classname.
> 
> If you'd rather not add this, that's fine.

I was just wanting to understand how you might trigger a tooltip mouse event in a print context as it's probably something I'm overlooking :).  I do think the CSS approach (`print:hidden`) would be a better approach since javascript doesn't always get an extra tick so `window.matchMedia()` might not get a chance to trigger.

PR #375, opened 2024-06-01
Detect package manager in create script to give customized next steps

ttytm:
Will print the next steps based on the used package manager. 
E.g. this will also include the new kid on the block _bun_:

Examples:
|||
|:-:|:-:|
|**npm**|**yarn**|
|Screenshot_20240601_030041|Screenshot_20240601_030422|
|**pnpm**|**bun**|
|Screenshot_20240601_031225|Screenshot_20240601_030509|

If this change can make it, is a changeset needed?

### DIFF
--- a/.changeset/rare-suits-fold.md
+++ b/.changeset/rare-suits-fold.md
@@ -1,0 +1,5 @@
+---
+'create-svelte-ux': minor
+---
+
+Detect package manager for customized "Next Steps"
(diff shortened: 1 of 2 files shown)


techniq:
@ttytm This is a nice improvement, thanks!

Changesets are always required to bump versions, and thus be deployable to NPM. Documentation only fixes are the only changes that do not need a changeset.

ttytm:
Understood

PR #377, opened 2024-06-01
Bump dependency versions in create templates

ttytm:


techniq (approved):

packages/create-svelte-ux/templates/starter/package.json
```diff
@@ -16,31 +16,31 @@
...
-    "svelte-check": "^3.4.3",
+    "@mdi/js": "^7.4.47",
+    "@playwright/test": "^1.44.1",
+    "@sveltejs/adapter-auto": "^3.2.1",
+    "@sveltejs/kit": "^2.5.10",
+    "@sveltejs/vite-plugin-svelte": "^3.1.1",
+    "@typescript-eslint/eslint-plugin": "^7.11.0",
+    "@typescript-eslint/parser": "^7.11.0",
+    "autoprefixer": "^10.4.19",
+    "eslint": "^8.57.0",
```
ttytm, line 27: Eslint is the only one down to a major 8. It's a conflicting dep in `"@typescript-eslint/parser"` and a major eslint `9` was running into not only warnings but errors when trying to use the updated templates with npm.

techniq: Thanks @ttytm.  I personally find eslint more of a pain than helpful most of the time (I don't use it in Svelte UX [directly](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/package.json) :).  No worries if it lags.

techniq:
It would be nice if these were picked up by `pnpm update -r -i --latest` at some point.  Not sure if we could add them to the [workspace](https://github.com/techniq/svelte-ux/blob/main/pnpm-workspace.yaml).  Regardless, lets merge this and the Svelte 5 CLI now :)

packages/create-svelte-ux/templates/starter/package.json
ttytm: > I personally find eslint more of a pain than helpful most of the time. 

Same experience :see_no_evil:

PR #391, opened 2024-06-13
Fix .contains() does not mean it's a child

myieye:
Three things were wrong here:
- The logic here wasn't quite right. `elem1.contains(elem2)` checks all descendants, so it doesn't imply that elem2 is a child of elem1
- We shouldn't recalculate the target when the element is getting destroyed, because we won't necessarily get the current target if there have been DOM changes
- `onDestroy` was using the original `options`, which would be out of date if they had ever been changed (i.e. via `update()`)

Due to a combination of these problems, when I navigate away from a route in my application I'm currently getting:
```
Uncaught (in promise) DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at destroy (http://localhost:5173/node_modules/.vite/deps/svelte-ux.js?v=ed6d092c:3364:16)
```

There are just so many little details that are wrong that it turns out there are a few ways to reproduce the bug, but what's crucial in my example is that:
1) The portal is initialized with `{target: undefined}` so that in `onDestroy` it queries for the nearest target.
2) There's a portal target (e.g. `.PortalTarget`) within the same `{#if}` as the portal content. Otherwise, `.contains()` will return false. The reason is that everything within the `{#if}` gets pulled out of the DOM, so root elements in the `{#if}` don't have a `parentElement`, but the elements within that detached subtree are still conntected to each other.
3) The portal content is not using that target i.e. it's not a direct child of it. It can simply be in its original parent (with `enabled: false`) or be in a different target that's also within the same `{#if}`. So, in the example you can trigger the error whether or not you've clicked "Move to target".

### DIFF
--- a/.changeset/ninety-seals-marry.md
+++ b/.changeset/ninety-seals-marry.md
@@ -1,0 +1,5 @@
+---
+"svelte-ux": patch
+---
+
+Fix destroyable portal logic
--- a/packages/svelte-ux/src/lib/actions/portal.ts
+++ b/packages/svelte-ux/src/lib/actions/portal.ts
@@ -21,37 +21,41 @@ export const portal: Action<HTMLElement, PortalOptions> = (node, options) => {
     fallbackTarget: node.closest('.PortalTarget') ?? document.querySelector('.PortalTarget'),
     originalParent: node.parentElement,
   };
-  moveNode(node, options, targets);
+  let currentTarget = moveNode(node, options, targets);
 
   return {
     update(options) {
-      moveNode(node, options, targets);
+      currentTarget = moveNode(node, options, targets);
     },
     destroy() {
-      const target = getTarget(options, targets.fallbackTarget);
       // If target still contains node that was moved, remove it.  Not sure if required
-      if (target?.contains(node)) {
-        target.removeChild(node);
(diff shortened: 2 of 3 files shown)


techniq:
@myieye This makes sense to me.  Could you add an example of this use case (nested portal'd element) and run `pnpm lint` to make prettier happy 😁.  Thanks!

myieye:
@techniq sorry, I jumped the gun on this one. But, I eventually nailed it down and added an example that breaks without my second commit.

techniq:
@myieye I really appreciate the detailed write up and digging in.  I've got 2 small asks :)

- I noticed you can not recreate the example after destroying due to the `{#if !destory}` around the whole example.  Can this be scoped further down, or at least put the Recreate button in an `{:else}` block of this statement?
- Could you add a changeset (`pnpm changeset`) so this creates a new version?

Thanks for all the effort here!

techniq:
@myieye Looks good!  Ready to merge?

myieye:
@techniq Yes 👍

techniq (approved):

techniq:
Thanks @myieye.  Publishing as `0.66.8`.  Should be live in a minute or so

PR #400, opened 2024-06-20
Fix scrolling away when opening menu

myieye:
I put a my reproduction from the recordings followed by the bug fix in [this branch](https://github.com/myieye/svelte-ux/tree/bug/focusMove-scrolls-to-parent-reproduction). (i.e. revert the fix to see the bug)

Before:
broken-scrolling

After:
fixed-scrolling

I'm not sure how to create an example for this, because I can only reproduce it when <body>/<html> are larger than the viewport. Somehow focusing on the element too early makes the browser scroll to the wrong place.

### DIFF
--- a/.changeset/swift-paws-dance.md
+++ b/.changeset/swift-paws-dance.md
@@ -1,0 +1,5 @@
+---
+"svelte-ux": patch
+---
+
+Fix scrolling away when opening menu
--- a/packages/svelte-ux/src/lib/actions/focus.ts
+++ b/packages/svelte-ux/src/lib/actions/focus.ts
@@ -1,5 +1,5 @@
-import { tick } from 'svelte';
 import type { ActionReturn } from 'svelte/action';
+import { delay } from '$lib/utils/promise.js';
 
 export function focusMove(
   node: HTMLElement,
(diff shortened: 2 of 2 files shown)


techniq:
Hmm, that's interesting, and fix works for me.  Someone reported a similar isolated issue on Discord and wonder if this was the cause as well.  Thanks!

I think the way AppLayout has `main` [overflow](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/AppLayout.svelte#L86) hides these kinds of issues a lot of times, and is something I would like to change.  Currently since main overflows, it breaks initial browser scrolling of link fragments (`#here`).

techniq:
@myieye Hmm, I just noticed, but it looks like this breaks the focus on QuickSearch.  Here's a [build](https://f5befa09.svelte-ux.pages.dev/) from yesterday that works, and reverting locally restores the behavior.

Currently [QuickSearch](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/QuickSearch.svelte#L24) uses [autoFocus](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/actions/input.ts#L6) which is an earlier version of `focusMove`.

Going to look into this, but wanted to let you know as I'll wait to cut a release (also looked at your other PR and was writing up some comments when I spotted this)

techniq:
Fixed via https://github.com/techniq/svelte-ux/commit/7d9f09cc999eea8d8244edffe818a36220680621.

I unified the implementations by calling `focusMove()` within `autoFocus()`.  I considered removing `autoFocus()` at first, but the best solution I could come up with was having a longer delay for those use case (`TextField` within `Dialog`, where Dialog is calling `focusMove()`).  I think having the distinction between `autoFocus` (used for inputs) and `focusMove` (more lower level, but mostly for modals/etc, makes sense.  What do you think?

Issue #246, opened 2024-02-09
Further unify `SelectField` and `MultiSelectField` options


techniq:
- [ ] MultiSelectField
  - [ ] [Change](https://github.com/techniq/svelte-ux/blob/next/packages/svelte-ux/src/lib/components/MultiSelectField.svelte#L31) `labelProp` default from `name` to `label`
- [ ] SelectField
  - [ ] [Align](https://github.com/techniq/svelte-ux/blob/next/packages/svelte-ux/src/lib/components/SelectField.svelte#L31-L32) `optionText()` / `optionValue()` with MultiSelectField's `labelProp` / `valueProp`
    - Do we even need this, or just require the processing it done in `options`?
- [ ] Align value types (`any` vs `string[]` / `number[]`).  Use generics
- [ ] Do the same in `MultiSelect`, `MultiSelectMenu` and `_SelectListOptions`

PR #401, opened 2024-06-20
Fix error from keying on stringified objects

myieye:
As the added example demonstrates, if you use objects in a SelectField, it causes an error, because the keyed each has duplicate keys (i.e. `[Object object]`). Perhaps I should just remove the `-${optionValue(option)}` part? In theory that would be a _slightly_ more breaking change, but it would probably be fine.

There's an inconsistency in the select-fields in that the MultiSelect value field is of type string[] | number[] 🤷 while SelectField value is of type `any`. Is that intentional?

### DIFF
--- a/.changeset/little-penguins-sort.md
+++ b/.changeset/little-penguins-sort.md
@@ -1,0 +1,5 @@
+---
+"svelte-ux": patch
+---
+
+Fix error from keying on stringified objects
--- a/packages/svelte-ux/src/lib/components/_SelectListOptions.svelte
+++ b/packages/svelte-ux/src/lib/components/_SelectListOptions.svelte
@@ -75,7 +75,7 @@
     onKeyPress(e);
   }}
 >
-  {#each filteredOptions ?? [] as option, index (`${option.group}-${optionValue(option)}`)}
+  {#each filteredOptions ?? [] as option, index (`${option.group}-${optionValue(option)}-${optionText(option)}`)}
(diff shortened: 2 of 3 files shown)


techniq:
> As the added example demonstrates, if you use objects in a SelectField, it causes an error, because the keyed each has duplicate keys (i.e. `[Object object]`).
>
> Perhaps I should just remove the `-${optionValue(option)}` part? In theory that would be a _slightly_ more breaking change, but it would probably be fine.

There is a use case where you can have the same value in multiple groups (albeit rare).  The selected value in the menu will also show as the first one (since it doesn't know which group was selected from), but this does require the fully qualified key (`group`/`value`).

image

Instead of including the label (since the values will be `[Object object]`), what if instead we `JSON.stringify(...)` the value.  This will technically be slower, but unlikely perceivable.

> There's an inconsistency in the select-fields in that the MultiSelect value field is of type string[] | number[] 🤷 while SelectField value is of type `any`. Is that intentional?

This is not intentionally.  There is a lot of alignment between the various select fields that need to be done (see https://github.com/techniq/svelte-ux/issues/246).  While a breaking change, I think it's worth getting rid of `optionText()`/`optionValue()` and `labelProps`/`valueProp` and require the user to pre-process the options into the standard `{ label: string, value: T }` structure, which will simplifying each implementation.  Align values and leveraging generics more is another big win.  These kind of changes are one of the reasons I'm keeping Svelte UX pre-1.0 so they can be made within running up the major version numbers.

PR #419, opened 2024-07-02
Pass object of classes through Select-like fields to Field instances

brandonmcconnell:
This PR does a few things
- adds a function `normalizeClasses` to normalize classes when they can be a string or an object, essentially converting strings to objects (e.g. `'p-2'` ➞ `{ root: 'p-2' }`)

  <table>
  <thead>
  <tr>
  <th>Before</th>
  </tr>
  </thead>
  <tbody>
  <tr></tr>
  <tr>
  <td>
                  
  ```svelte
  <Component classes={typeof className === 'string' ? { root: className } : className} />
  ```
  
  </td>
  </tr>
  </tbody>
  <thead>
  <tr>
  <th>After</th>
  </tr>
  </thead>
  <tbody>
  <tr></tr>
  <tr>
  <td>
                  
  ```svelte
  <Component classes={normalizeClasses(className)} />
  ```
                  
  </td>
  </tr>
  </tbody>
  </table>

- adds a function `clsMerge` for merging groups of classes without needing to do more complex property-by-property merging

  <table>
  <thead>
  <tr>
  <th>Before</th>
  </tr>
  </thead>
  <tbody>
  <tr></tr>
  <tr>
  <td>
                  
  ```svelte
  <Component classes={{
    root: cls(settingsClasses.root, 'grid', classes.root),
    field: cls(settingsClasses.field, 'flex', classes.field)
  }} />
  ```
  
  </td>
  </tr>
  </tbody>
  <thead>
  <tr>
  <th>After</th>
  </tr>
  </thead>
  <tbody>
  <tr></tr>
  <tr>
  <td>
                  
  ```svelte
  <Component classes={clsMerge(settingsClasses, { root: 'grid', field: 'flex' }, classes)} />
  ```
                  
  </td>
  </tr>
  </tbody>
  </table>

- passes the `classes` objects down to the `TextField` component from the `Select` and `MultiSelect` components
- fixes a TypeError caused by a faulty ReturnType from `entries()` and removes a related `@ts-expect-error`

PR #421, opened 2024-07-02
Patch incorrect order of classnames (regression from #419)

brandonmcconnell:
This resolves a regression re incorrect classname order, caught in this comment:
https://github.com/techniq/svelte-ux/pull/419#issuecomment-2204019023


techniq:
Thanks @brandonmcconnell. This strategy will be good to apply throughout as we get time.

PR #423, opened 2024-07-04
Update TS type for `TreeNode` to be recursive

brandonmcconnell:
~This is a small TS type fix, so I don't think a changeset is necessary, but let me know if you feel otherwise.~

I went ahead and added a changeset after all, but feel free to remove it.

### DIFF
--- a/packages/svelte-ux/src/lib/utils/array.ts
+++ b/packages/svelte-ux/src/lib/utils/array.ts
@@ -188,7 +188,7 @@ export function nestedFindByPredicate(
   return undefined;
 }
 
-export type TreeNode = { id: string; name: string; level: number; children: Node[] };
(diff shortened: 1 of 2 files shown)


packages/svelte-ux/src/lib/utils/array.ts
```diff
@@ -188,7 +188,7 @@ export function nestedFindByPredicate(
   return undefined;
 }
 
-export type TreeNode = { id: string; name: string; level: number; children: Node[] };
+export type TreeNode = { id: string; name: string; level: number; children: TreeNode[] };
```
brandonmcconnell, line 191: I think this was meant to be `TreeNode`, or that the proper TS type for `Node` wasn't imported, as it currently appears to be using the browser-default `Node` type.

techniq (approved):
Thanks @brandonmcconnell

PR #425, opened 2024-07-08
[SelectField] Fix options with object value due to non-unique {#each} keys

techniq:

### DIFF
--- a/.changeset/tough-actors-switch.md
+++ b/.changeset/tough-actors-switch.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+[SelectField] Fix options with object value due to non-unique {#each} keys
--- a/packages/svelte-ux/src/lib/components/_SelectListOptions.svelte
+++ b/packages/svelte-ux/src/lib/components/_SelectListOptions.svelte
@@ -68,7 +68,7 @@
     onKeyPress(e);
   }}
 >
-  {#each filteredOptions ?? [] as option, index (`${option.group}-${option.value}`)}
+  {#each filteredOptions ?? [] as option, index (JSON.stringify(option))}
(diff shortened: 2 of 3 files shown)


techniq:
Replaces #401.  cc: @myieye

myieye (approved):
@techniq sorry I didn't reply to your questions a few weeks back. I'm on parental leave now.

Your solution is a bit cleaner than mine. I'm fine with it. I think JSON.stringify is quite fast.

Though, personally I think I'd still choose mine. I'm using short lists of medium sized objects and long lists of small objects. But there's nothing preventing someone from using long lists of large objects and then this could add a bit of unnecessary lag.

techniq:
First, congrats on the parental leave, and you're doing it wrong if you're responding to Github PRs :).  I hope baby and mommy are doing good.

I agree on the performance on large lists might be measurable, and we'll address that if needed.  I was concerned there might be cases where labels are not unique, and while groups might be leveraged, there is no guarantee.  You can also attach additional [properties](https://github.com/techniq/svelte-ux/blob/592cbbb55ebb20edd9f213863cf93f616eed774c/packages/svelte-ux/src/lib/types/index.ts#L12) to a menu option so `JSON.stringify(option)` the full option increases the likelihood of uniqueness across options.

I've also exposed a `getKey()` prop in the past if you wanted more control, which might be a solution if JSON.stringify performance is an issue.  Also, I like to expose full control in a default slot / children so the user is able to implement their own `{#each}` block with more control for transitions, etc (something I do in [LayerChart](https://github.com/techniq/layerchart/commit/5a57e6353feb3292e776283817c15b4e3a6b2307) a good bit).

Anyways, I'm going to ship this approach for now, but will keep an eye on performance regressions.

PR #431, opened 2024-07-10
Add `MultiSelect` fallback text when `maintainOrder` no filtered items

brandonmcconnell:
Work performed outlined [below](https://github.com/techniq/svelte-ux/pull/431#discussion_r1671385844).


techniq:
Thanks @brandonmcconnell

Issue #144, opened 2023-11-29
Svelte 5 compatibility


techniq:
Issue to track any Svelte 5 compatibility issues.  Hoping close to `0` changes are required in Svelte UX to support 5.

---
## ⚠️ PENDING: sveld (API docs) not compatible
issue: #435 

---
## ✅ FIXED: `MultiSelect` hangs browser (MultiSelectField / MultiSelectMenu appear to work fine)
example: https://svelte5-docs.svelte-ux.pages.dev/docs/components/MultiSelect (see [screenshot](https://github.com/techniq/svelte-ux/issues/144#issuecomment-2261124165))
fix: https://github.com/techniq/svelte-ux/issues/507

---

## ✅ FIXED: `SelectField` raises `node_invalid_placement_ssr`
Server-side
> node_invalid_placement_ssr: `<button>` (src/lib/components/SelectField.svelte:415:0) cannot contain `<button>` (src/lib/components/Button.svelte:446:0)
>
> This can cause content to shift around as the browser repairs the HTML, and will likely result in a `hydration_mismatch` warning.

Client-side
> Uncaught (in promise) HierarchyRequestError: Failed to execute 'appendChild' on 'Node': This node type does not support this method.

fix: https://github.com/techniq/svelte-ux/pull/436/commits/4e172184edd336e445f4265331efc0abb7530d7d

---
## ✅ FIXED: `Form` `$: dispatch('change', ...)` called immediately (not after change)
examples: [svelte 4](https://svelte.dev/repl/992a3d0077af44819191534599ca7966?version=4.2.18), [svelte 5](https://svelte-5-preview.vercel.app/#H4sIAAAAAAAAE5WQQWuEMBCF_8owFFxBtOdUF0rbX1F7CHHcDcQkJKNQJP-9ZFe0tL30MId5ee8xX1YctaGI4n1FKydCgc_eY4X86fMSFzJMWGF0c1BZaaMK2vO5tz3rybvA8OIm7yxZhjG4CYq62ZX6XlA89bZtjqRtj4yzQl2lvVC3EnRnUM5GZ6g27nIq7i9FBVQPxFKbMkFza8AKJzfoUdOAgsNMqdoR9vJ_gqygAkmmt4Usv-roJasrBUgb2AGTQ_lQhmGzQfd3-FRmOxtiWKSZCTp4vCl5HsSe_8a6bs5U_vy239Af6QupmNI8wwEAAA==)
[...]

PR #447, opened 2024-07-31
Avoid uninitialized variable error in Duration component with Svelte 5

dimfeld:
related to #144


techniq:
Much appreciated @dimfeld!  I tried fixing this using `import { get } from 'svelte/store'` but that didn't work.  I appreciate the [explanation](https://github.com/techniq/svelte-ux/issues/144#issuecomment-2259709984) as well

dimfeld:
Yeah, using get is sort of what the svelte 5 code is doing under the hood anyway.

PR #448, opened 2024-07-31
Don't use `bind:value` on SelectField that has no options

dimfeld:
Fixes SelectFIeld issue in #144


techniq:
Thanks for the fix and [explanation](https://github.com/techniq/svelte-ux/issues/144#issuecomment-2259694104)

PR #452, opened 2024-08-04
Cleanup comments and conditions

baterson:
Hello, this PR removes some comments and unnecessary condition branching


techniq:
Thanks @baterson.  Some of these comments might be useful as `logger.debug()` in the future, but no reason to keep them for now.

techniq (approved):

PR #467, opened 2024-08-16
Correct `on:close` usage in Dialogs and Drawers where persistent

brandonmcconnell:
Correct docs to match the changes from #464

### DIFF
--- a/packages/svelte-ux/src/routes/docs/components/Dialog/+page.svelte
+++ b/packages/svelte-ux/src/routes/docs/components/Dialog/+page.svelte
@@ -69,7 +69,7 @@
 <Preview>
   <Toggle let:on={open} let:toggleOn let:toggleOff>
     <Button on:click={toggleOn}>Show Dialog</Button>
-    <Dialog {open} on:close={toggleOff} {loading} persistent={loading}>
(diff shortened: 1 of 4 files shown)


techniq:
@brandonmcconnell I'm starting to change my mind back to dispatching a separate `closeAttempt` event when `persistent` is set and a close is attempted, instead of having to check `e.detail.open` on the `close` event.  This will simplify most of the common `persistent` use cases (i.e. revert most of this PR) and still provide a hook to prompt a user for confirmation.

- [Dialog](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Dialog.svelte#L65-L66)
- [Drawer](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Drawer.svelte#L41-L42)

The refinements done in #464, including a common `close()` method exposed as a slot prop and called consistently, as well as the introduction of the `close({ force: true })` were great improvements, but I think the single `close` event is more error prone.

I forgot when we chatted if there was a big reason to have a single `close` event for both cases (as I know we went through a few iterations on that side).

techniq:
Closed in favor of #468

PR #476, opened 2024-09-05
Dispatch `apply` event when applying from MultiSelect

brandonmcconnell:


techniq (approved):
LGTM.  Thanks @brandonmcconnell

PR #483, opened 2024-09-12
Allow `onApply` to accept either a synchronous or asynchronous function

brandonmcconnell:
- Introduce `MaybePromise` and `MaybeAsync` type helpers
- Allow `onApply` to accept either a synchronous or asynchronous function

### DIFF
--- a/.changeset/blue-vans-whisper.md
+++ b/.changeset/blue-vans-whisper.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+Introduce `MaybePromise` and `MaybeAsync` type helpers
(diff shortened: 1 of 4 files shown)


techniq (approved):
Care to also PR this to LayerStack, which is where 99% of Svelte UX's utils, actions, stores, etc will be living going forward.

I've extracted them as separate packages, and the LayerChart will soon only have a devDependency on Svelte UX after simplified [PRs](https://github.com/techniq/layerchart/pull/224) merge (LayerStack will also be getting a lot of utils from LayerChart for a common location that can be used in non-Svelte UX/LayerChart/and some cases Svelte projects.

Svelte UX will be updated to use `@layerstack/*` packages for most/all of these, and likely re-export to reduce the breaking change until closer to 1.0 (still debating)

techniq:

PR #484, opened 2024-09-13
Expose `tabIndex` on `Button` and `MenuItem`

brandonmcconnell:


packages/svelte-ux/src/lib/components/Button.svelte
```diff
@@ -448,10 +449,11 @@
   {href}
   {target}
   {type}
+  {disabled}
   {...$$restProps}
   class={_class}
   style={$$props.style ?? ''}
-  {disabled}
+  tabindex={tabIndex}
```
techniq, line 456: Hmm... technically this is exposed via `...$$restProps`, although I know intellisense doesn't really work out nice here.  Any reason to add it explicitly (and as camel case vs the DOM `tabindex`)?

I have hopes to improve this with Svelte 5 using something like...
```js
import type { HTMLButtonAttributes } from 'svelte/elements';

interface MyProps extends HTMLButtonAttributes  {
	required: string;
	optional?: number;
	partOfEverythingElse?: boolean;
};
let { required, optional, ...everythingElse }: MyProps = $props();
```

I've been using...
```ts
interface $$Props extends ComponentProps<Chart<TData>> {
  series?: typeof series;
  labels?: typeof labels;
  points?: typeof points;
  props?: typeof props;
  stackSeries?: typeof stackSeries;
}
```
a little in the Simplified Charts [PR](https://github.com/techniq/layerchart/pull/224/files#diff-027fff683d6eef2f5e3c2c3ae7195529c25a7870598bc78233c75ed2fd832774R19-R25) but it's also painful to double declare each prop, so would like to wait til Svelte 5 and $props rune before going too far down that path.

techniq:
Going to close for now as `tabindex` is already available via`{...$$restProps}` and hopefully intellisense will improve after the migration to Svelte 5

PR #487, opened 2024-09-21
Add basic Toast component

willm0602:
Adds a basic version of a toast component that wraps Notifications and includes them on the bottom

Wanted to also leave a couple of notes before this gets merged
- I wasn't able to find any tests for components specifically in the codebase. Is there anywhere I can add these before we merge these?
- When I was testing this out, I wasn't able to initially since I got an issue, `'TZ' is not recognized as an internal or external command, operable program or batch file`. I modified the test script in package.json to be `SET TZ=UTC+4 vitest` and that let the test suite run, but failed on six-issues, not sure if it's an issue with developing on windows OS

chrome_7oGfLfmPJR

### DIFF
--- a/packages/svelte-ux/src/lib/components/Toast.svelte
+++ b/packages/svelte-ux/src/lib/components/Toast.svelte
@@ -1,0 +1,14 @@
+<script lang="ts">
+  import toastStore from '$lib/stores/toastStore.js';
+  import Notification from './Notification.svelte';
+</script>
+
+<div class="fixed bottom-0 left-1/2 translate-x-[-50%] z-50 gap-y-4">
+  {#each $toastStore as toast}
+  <div class="mt-4">
+    <Notification open={true} closeIcon>
+      <span slot="title">{toast.text}</span>
+    </Notification>
+  </div>
+  {/each}
+</div>
--- a/packages/svelte-ux/src/lib/components/Toast.svelte
+++ b/packages/svelte-ux/src/lib/components/Toast.svelte
@@ -1,14 +1,17 @@
 <script lang="ts">
-  import toastStore from '$lib/stores/toastStore.js';
+  import toastStore, { removeToast } from '$lib/stores/toastStore.js';
   import Notification from './Notification.svelte';
+  import {portal} from '../actions/portal.js';
 
   import {cls} from '../utils/styles.js';
+  import Button from './Button.svelte';
 
   type Placement = 'top' | 'top-right' | 'right' | 'bottom-right' |
                    'bottom' | 'bottom-left' | 'left' | 'top-left';
 
   const DEFAULT_PLACEMENT: Placement = 'bottom';
   export let placement: Placement = DEFAULT_PLACEMENT;
+  export const classes: string = '';
 
   const isOnTop = placement.includes('top');
   const isOnBottom = placement.includes('bottom');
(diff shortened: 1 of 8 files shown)


techniq:
> I wasn't able to find any tests for components specifically in the codebase. Is there anywhere I can add these before we merge these?

I currently only have `vitest` setup for unit tests (mostly utils, example [format.test.ts](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/utils/format.test.ts)).

I would like to setup playwright for integration/browser tests and integrated into the CI pipeline, similar to shadcn-svelte and melt-ui's setup, but haven't had the time to tackle it yet.  If that is an initiative you would like to get started, I would kindly review a PR 😁

> When I was testing this out, I wasn't able to initially since I got an issue, 'TZ' is not recognized as an internal or external command, operable program or batch file. I modified the test script in package.json to be SET TZ=UTC+4 vitest and that let the test suite run, but failed on six-issues, not sure if it's an issue with developing on windows OS

Sadly most of the development occurs on Mac and thus a unix environment (and usually zsh or bash terminal).  With that said, one of our contributors was using Windows and leveraging the [WSL](https://learn.microsoft.com/en-us/windows/wsl/) (he actually was the one to add the timezone environment variable, although he also switched to Mac later on).

I don't hear of much Windows open source development happening without using WSL nowadays, but I'm also pretty detached from that ecosystem (switched to Mac 10+ years ago).

If you're not using WSL, I would recommend [setting](https://learn.microsoft.com/en-us/windows/wsl/install) it up and see if you have a better experience.

packages/svelte-ux/src/lib/components/Toast.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+  import toastStore from '$lib/stores/toastStore.js';
+  import Notification from './Notification.svelte';
+</script>
+
+<div class="fixed bottom-0 left-1/2 translate-x-[-50%] z-50 gap-y-4">
```
techniq, line 6: I could see us adding a `placement` prop to control which of the 8 corners/edges to place the notifications (at least `top-left`, `top`, `top-right`, `bottom-left`, `bottom`, `bottom-right`).

you can see some examples of this in [Drawer](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Drawer.svelte#L76-L83) and [Badge](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Badge.svelte#L12)

techniq: It would also be good if `Toast` supported class overrides, either directly (`<Toast classes={...}>`) or via [settings](https://svelte-ux.techniq.dev/customization#settings).

To support this, you would need to add a `classes` prop and also getting the classes defined in `settings(...)` using...

```svelte
<script>
  ...

  export let classes: {
    root?: string;
    actions?: string;
    // ...any other internal element/component you might want to pass a class to
  } = {};
  const settingsClasses = getComponentClasses('Toast');
</script>
```

and change...

```svelte
<div class="fixed bottom-0 left-1/2 translate-x-[-50%] z-50 gap-y-4">
```

to

```svelte
<div
    class={cls(
      'Toast',
      'fixed bottom-0 left-1/2 translate-x-[-50%] z-50 gap-y-4',
      settingsClasses.root,
      classes.root,
      $$props.class
    )}
>

techniq (commented):
@willm0602 Thanks for the PR!

I left some notes, some of which might not be worth addressing in this PR, but thought it was helpful to put my thoughts down to see what future improvements, etc.

Other ideas we chatted about in Discord (putting here to more easily track) were:

- When you `pointerover` a notification, we should stop the timeout and on `pointerout` we start/resume in case you are interacting with it (or just needed more time to read the message
- Consider integrating Progress / ProgressCircle to show the countdown (probably best when we add the `variant` support)

techniq:
Few other ideas I forgot to mention.  It might make sense to put an instance of the `toastStore` in [settings()](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/settings.ts#L119-L130) like we do for themes, locales, nav drawer, etc.  We could also consider adding an instance of `<Toast>` within `<AppLayout>` wired up to this instance.

willm0602:
@techniq started working on the changes and got an issue where I can't get the toasters to appear over the elements of the app layout that used fixed position (so the side and topnav) even after adjusting the z-indexes

image

Was wondering if you had any ideas for fixing this?

techniq:
> @techniq started working on the changes and got an issue where I can't get the toasters to appear over the elements of the app layout that used fixed position (so the side and topnav) even after adjusting the z-indexes
> 
> image
> Was wondering if you had any ideas for fixing this?

Modal elements like [Drawer](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Drawer.svelte#L22) and [Dialog](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/Dialog.svelte#L24) "portal" their content to the bottom of `<body>` via the [portal](https://svelte-ux.techniq.dev/docs/actions/portal) action and doing similar within Notification should solve the issue.

techniq:
@willm0602 I started work on improving/simplifying the `Notification` component by support props (not requiring but still supporting slots) such as `<Notification title="..." description="..." actions={...}>` and also supporting `color`/`variant`/etc props.  Take a look at the PR [preview](https://techniq-notification-variant.svelte-ux.pages.dev/docs/components/Notification).

I also have a [PR](https://github.com/techniq/svelte-ux/pull/489) to use `@layerstack/*` packages for most all svelte stores/actions/etc.  I'll plan to merge these hopefully in the next few days (the later is a much bigger change).

This will impact your PR some, but not sure where `toastStore` should live just yet (it's rather tied to Svelte UX Toast/Notification and not sure it makes much sense to be within `@layerstack/*`.  In some ways I think of it similar to AppBar/AppLayout/NavItem and their use of `showDrawer` within `settings`.

techniq:

PR #493, opened 2024-10-05
Fix Form server submit

techniq:

### DIFF
--- a/.changeset/small-pens-love.md
+++ b/.changeset/small-pens-love.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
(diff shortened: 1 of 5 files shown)


kasleet:
@techniq Thanks for the quick response and PR! I tried it and it works as expected, but I think this somehow broke the client-side form handling? At least for me using 0.75.0, the state is not updated anymore, using 0.74.9 works as expected. I would exclude the possibility that I have some bug in my code, but maybe you can have a look, to make sure its still working?

kasleet:
I created a PR which fixes the problem, but I don't know if you like this structure: https://github.com/techniq/svelte-ux/pull/494

PR #494, opened 2024-10-06
Update form stores correctly

kasleet:
https://github.com/techniq/svelte-ux/pull/493 introduced a problem with client-side form handling when using zod schemas - this fixes the issue.

### DIFF
--- a/packages/svelte-ux/src/lib/stores/formStore.ts
+++ b/packages/svelte-ux/src/lib/stores/formStore.ts
@@ -51,7 +51,6 @@ export default function formStore<T extends Objectish = any>(
           // Clear errors
           errorsStore.set({});
           // TODO: Consider using `result.data` in case there are defaults, etc?
-          return true;
         } else {
           const errors = {};
           for (const issue of result.error.issues) {
--- a/.changeset/weak-pandas-flow.md
+++ b/.changeset/weak-pandas-flow.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+Fix form store updates when using zod schema


package-lock.json
techniq, line 1: Could you remove `package-lock.json`.  This project uses `pnpm` by default.  Thanks!

packages/svelte-ux/src/lib/stores/formStore.ts
```diff
@@ -62,12 +62,14 @@ export default function formStore<T extends Objectish = any>(
         }
       }
 
-      const newState = finishDraft(draft, (patches, inverseChanges) => {
-        undoList.push(inverseChanges);
-      }) as T;
+      if (!options?.isFormSubmission) {
```
techniq, line 65: I'm trying to think this through, but do think not updating the local state when submitting the form is needed (or even desired).  It seems like it could flash an optimistic view while the form submission in inflight, but not sure if that's truly possible, or unwanted, but I could completely be overlooking something as well.

If we don't need this, it greatly simplifies everything as we only need to remove the early `return true` I added in the previous PR (I completely overlooked the later `return true` at the bottom of `commit()`.

Thoughts?

techniq (changes requested):
Thanks so much for the quick PR (sorry for the regression).

packages/svelte-ux/src/lib/stores/formStore.ts
kasleet: I also don't think that this would be an issue. I thought you added the early `return true` for performance reasons, but I also don't see a problem if we just update the store even though it might be not necessary.

techniq (approved):

techniq:
Thanks again for the PR.  Deployed as [0.75.1](https://github.com/techniq/svelte-ux/releases/tag/svelte-ux%400.75.1)

PR #505, opened 2024-10-30
Update Layerchart templates to match svelte-ux changes

northprint:
Layerchart layout provided in the template was broken
The reason was that it was not compatible with svelte-ux changes, so it has been fixed.

### before
スクリーンショット 2024-10-23 13 55 45

### after
スクリーンショット 2024-10-29 4 17 08

### DIFF
--- a/packages/create-svelte-ux/templates/layerchart/src/routes/+layout.svelte
+++ b/packages/create-svelte-ux/templates/layerchart/src/routes/+layout.svelte
@@ -5,13 +5,17 @@
 	import '../app.postcss';
 
 	settings({
-		classes: {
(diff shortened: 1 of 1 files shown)


techniq (approved):
Thanks @northprint!  I need to setup some CI/CD deployments to make sure these stay in sync

techniq:
@northprint I checked the [starter template](https://github.com/techniq/svelte-ux/blob/main/packages/create-svelte-ux/templates/starter/src/routes/%2Blayout.svelte) and it's already updated.  Thanks again

northprint:
You're welcome, glad I could help.

PR #525, opened 2024-12-15
feat(ThemeSwitch) Allow custom classes for icon

gavingolden:
### Changes
- adds the ability to pass custom classes to the underlying `Icon` components in `ThemeSwitch`.

Currently the `ThemeSwitch` icons are hard-coded to `text-primary` which makes it difficult to see the icons when using certain backgrounds (eg if my top nav bar uses the `secondary` color palette`)

Side note: This whole library is awesome and very easy to use/understand!

### DIFF
--- a/.changeset/beige-pugs-boil.md
+++ b/.changeset/beige-pugs-boil.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': minor
+---
+
+Allow custom styling of ThemeSwitch component
--- a/packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
+++ b/packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
@@ -5,8 +5,16 @@
   import Icon from './Icon.svelte';
 
   import { getSettings } from './settings.js';
+  import { cls } from '../utils/index.js';
+  import type { ComponentProps } from 'svelte';
 
   const { currentTheme } = getSettings();
+
+  export let classes: {
+    icon?: string;
+  } & ComponentProps<Switch>["classes"] = {};
+
+
 </script>
 
 <Switch
@@ -26,12 +34,18 @@
     <Icon
       data={mdiWeatherNight}
       size=".8rem"
-      class="row-[1] col-[1] text-primary opacity-0 dark:opacity-100"
+      class={cls(
+        "row-[1] col-[1] text-primary opacity-0 dark:opacity-100",
+        classes.icon,
+      )}
     />
     <Icon
       data={mdiWhiteBalanceSunny}
       size=".8rem"
-      class="row-[1] col-[1] text-primary opacity-100 dark:opacity-0"
+      class={cls(
+        "row-[1] col-[1] text-primary opacity-100 dark:opacity-0",
+        classes.icon,
+      )}
     />
   </div>
 </Switch>
--- a/packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
+++ b/packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
@@ -4,11 +4,15 @@
   import Switch from './Switch.svelte';
   import Icon from './Icon.svelte';
 
-  import { getSettings } from './settings.js';
+  import { getComponentSettings, getSettings } from './settings.js';
   import { cls } from '../utils/index.js';
   import type { ComponentProps } from 'svelte';
+  import { clsMerge } from '$lib/utils/styles.js';
 
   const { currentTheme } = getSettings();
+
+  const { classes: settingsClasses } = getComponentSettings('ThemeSwitch');
+
 
   export let classes: {
     icon?: string;
@@ -22,22 +26,22 @@
     let newTheme = e.target?.checked ? 'dark' : 'light';
     currentTheme.setTheme(newTheme);
   }}
-  classes={{
+  classes={clsMerge({
     switch: 'dark:bg-primary dark:border-primary',
     toggle: 'translate-x-0 dark:translate-x-full',
-  }}
+  }, settingsClasses.root, classes)}
   {...$$restProps}
 >
   <div class="grid grid-cols-1 grid-rows-1">
     <Icon
       data={mdiWeatherNight}
       size=".8rem"
-      class={cls('row-[1] col-[1] text-primary opacity-0 dark:opacity-100', classes.icon)}
+      class={cls('row-[1] col-[1] text-primary opacity-0 dark:opacity-100', settingsClasses.icon, classes.icon)}
     />
     <Icon
       data={mdiWhiteBalanceSunny}
       size=".8rem"
-      class={cls('row-[1] col-[1] text-primary opacity-100 dark:opacity-0', classes.icon)}
+      class={cls('row-[1] col-[1] text-primary opacity-100 dark:opacity-0', settingsClasses.icon, classes.icon)}
     />
   </div>
 </Switch>
--- a/packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
+++ b/packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
@@ -12,9 +12,7 @@
 
   export let classes: {
     icon?: string;
-  } & ComponentProps<Switch>["classes"] = {};
-
-
+  } & ComponentProps<Switch>['classes'] = {};
 </script>
 
 <Switch
@@ -34,18 +32,12 @@
     <Icon
       data={mdiWeatherNight}
       size=".8rem"
-      class={cls(
-        "row-[1] col-[1] text-primary opacity-0 dark:opacity-100",
-        classes.icon,
-      )}
+      class={cls('row-[1] col-[1] text-primary opacity-0 dark:opacity-100', classes.icon)}
     />
     <Icon
       data={mdiWhiteBalanceSunny}
       size=".8rem"
-      class={cls(
-        "row-[1] col-[1] text-primary opacity-100 dark:opacity-0",
-        classes.icon,
-      )}
+      class={cls('row-[1] col-[1] text-primary opacity-100 dark:opacity-0', classes.icon)}
     />
   </div>
 </Switch>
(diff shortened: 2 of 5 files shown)


gavingolden:
@techniq  I'm unfamiliar with changesets and am getting an error that I can't resolve. Lmk if you have any tips. I ran `npm i` and the changesets command from the project root.
```
> npm run changeset

> ...(do minor bump on svelte-ux)

🦋Note: All dependents of these packages that will be incompatible with the new version will be patch bumped when this changeset is applied.
🦋
🦋  Is this your desired changeset? (Y/n) · true
🦋  error Error: Cannot find module 'prettier-plugin-svelte'
🦋  error Require stack:
🦋  error - /Users/me/code/me/svelte-ux/node_modules/prettier/index.js
🦋  error - /Users/me/code/me/svelte-ux/node_modules/@changesets/write/dist/changesets-write.cjs.js
🦋  error - /Users/me/code/me/svelte-ux/node_modules/@changesets/cli/dist/changesets-cli.cjs.js
🦋  error - /Users/me/code/me/svelte-ux/node_modules/@changesets/cli/bin.js
🦋  error     at Function._resolveFilename (node:internal/modules/cjs/loader:1249:15)
🦋  error     at resolve (node:internal/modules/helpers:151:19)
🦋  error     at /Users/me/code/me/svelte-ux/node_modules/prettier/index.js:38119:25
🦋  error     at Array.map (<anonymous>)
🦋  error     at Object.load (/Users/me/code/me/svelte-ux/node_modules/prettier/index.js:38114:65)
🦋  error     at Object.load [as loadPlugins] (/Users/me/code/me/svelte-ux/node_modules/prettier/index.js:16147:23)
🦋  error     at /Users/me/code/me/svelte-ux/node_modules/prettier/index.js:38181:24
🦋  error     at Object.format (/Users/me/code/me/svelte-ux/node_modules/prettier/index.js:38197:12)
🦋  error     at Object.writeChangeset [as default] (/Users/me/code/me/svelte-ux/node_modules/@changesets/write/dist/changesets-write.cjs.js:104:26)
🦋  error     at async add (/Users/me/code/me/svelte-ux/node_modules/@changesets/cli/dist/changesets-cli.cjs.js:542:25) {
🦋  error   code: 'MODULE_NOT_FOUND',
🦋  error   requireStack: [
🦋  error     '/Users/me/code/me/svelte-ux/node_modules/prettier/index.js',
🦋  error     '/Users/me/code/me/svelte-ux/node_modules/@changesets/write/dist/changesets-write.cjs.js',
🦋  error     '/Users/me/code/me/svelte-ux/node_modules/@changesets/cli/dist/changesets-cli.cjs.js',
🦋  error     '/Users/me/code/me/svelte-ux/node_modules/@changesets/cli/bin.js'
🦋  error   ]
🦋  error }
```

techniq:
@gavingolden Thanks for the PR and kind words.

A few ideas to take this even further (if you want).

Svelte UX has the concept of [settings](https://svelte-ux.techniq.dev/customization#settings) which allows you to apply class overrides at a global/context level.  To enable this for a component, simply read the components classes from [getComponentSettings()](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/SelectField.svelte#L26) and [apply](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/SelectField.svelte#L421) it using `cls()` (like you are for component-level `classes` prop).  `settingsClasses` should come before component `classes` to allow the to be overridden on a per-instance basis (i.e. using `<ThemeSwitch classes={...}`>`

Second, along with the `Icon` classes, we could also support passing [classes](https://github.com/techniq/svelte-ux/pull/525/files#diff-056cd0d633991afdabe014692fdb8d1505a20bb6242856eb87d9eb861bdd7ffdR25-R28) to the `Switch` component within ThemeSwitch (ex. change the background or border color).  Since `<Switch classes={{...}}>` is an object of inner elements, you can handle this using the [clsMerge](https://github.com/techniq/svelte-ux/blob/e1fc9305ebe8168fbc1cfaee20561163a19a9576/packages/svelte-ux/src/lib/utils/styles.ts#L49-L52) util.  You can see [SelectField](https://github.com/techniq/svelte-ux/blob/main/packages/svelte-ux/src/lib/components/SelectField.svelte#L447-L456) or [MultiSelectField](https://github.com/techniq/svelte-ux/blob/e1fc9305ebe8168fbc1cfaee20561163a19a9576/packages/svelte-ux/src/lib/components/MultiSelectField.svelte#L163-L167) for examples.  This would allow `<ThemeSwitch classes={{ switch: { ... } }}>`

As for `changesets`, I wonder if this is because the project is using `pnpm` workspaces.  Can you install `pnpm` and run `pnpm i` from the project, then try `pnpm changeset` from the root of the project?

Thanks again for the PR!

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
```diff
@@ -4,9 +4,18 @@
...
 
-  import { getSettings } from './settings.js';
+  import { getComponentSettings, getSettings } from './settings.js';
+  import { cls } from '../utils/index.js';
+  import type { ComponentProps } from 'svelte';
+  import { clsMerge } from '$lib/utils/styles.js';
 
   const { currentTheme } = getSettings();
+
+  const { classes: settingsClasses } = getComponentSettings('ThemeSwitch');
```
gavingolden, line 14: Not sure if there's a meaningful difference between `getComponentSettings` and `getComponentClasses` in this case

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
```diff
@@ -4,9 +4,18 @@
...
+
+  export let classes: {
+    icon?: string;
+  } & ComponentProps<Switch>['classes'] = {};
```
gavingolden, line 18: `classes.root` would actually be passed down as the `root` for the underlying `Switch` which might be slightly confusing but I think it's fine

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
```diff
@@ -16,22 +25,34 @@
...
-    switch: 'dark:bg-primary dark:border-primary',
-    toggle: 'translate-x-0 dark:translate-x-full',
-  }}
+  classes={clsMerge(
+    {
+      switch: 'dark:bg-primary dark:border-primary',
+      toggle: 'translate-x-0 dark:translate-x-full',
+    },
+    settingsClasses.root,
+    classes
```
gavingolden, line 34: it isn't explicit here but if someone did `<ThemeSwitch class="..." />` then the `class` would still get passed down to the underlying `Switch` via `$$restProps`

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
```diff
@@ -16,22 +25,34 @@
...
   {...$$restProps}
 >
   <div class="grid grid-cols-1 grid-rows-1">
     <Icon
       data={mdiWeatherNight}
       size=".8rem"
-      class="row-[1] col-[1] text-primary opacity-0 dark:opacity-100"
+      class={cls(
+        'row-[1] col-[1] text-primary opacity-0 dark:opacity-100',
+        settingsClasses.icon,
```
gavingolden, line 44: is it correct that this should inherit the `ThemeSwitch`-specific icon classes or should it instead be the global `Icon` classes? Or a combo of the two? 🤔

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
techniq: Awe, good point.  `getComponentSettings` is only useful when there are other settings/defaults to read (`variant`,  `labelPlacement`, etc).

With that said, feel free to use `const settingsClasses = getComponentClasses('ThemeSwitch');` to streamline this a bit.

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
techniq: that's correct

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
techniq: it should be the `ThemeSwitch` icon specific case here, but `Icon` settings will also apply "globally" to all `Icon` instances (direct or within other components such as here within ThemeSwitch).

.changeset/beige-pugs-boil.md
techniq, line 1: 🎉

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
techniq: This is interesting as I was initially expecting adding `switch` to `classes` here, but `classes.switch.switch` is a little odd so handling it this way seems fine (just different than normal)

techniq (approved):
LGTM.  Let me know if you want to change `getComponentSettings` to `getComponentClasses`, but either way is fine with me.  Thanks again 🫶

gavingolden:
@techniq  thanks `pnpm` did the trick for the changset and I made the other improvements. I tested the settings with this  in the component preview page (which I didn't commit). Let me know if you see any other potential improvements, otherwise I think this is ready to go
```svelte
<script lang="ts">
  import { settings, ThemeSwitch } from 'svelte-ux';
  import Preview from '$lib/components/Preview.svelte';

  settings({
    components: {
      ThemeSwitch: {
        classes: {
          root: 'border',
          icon: 'text-primary-content',
          switch: 'bg-secondary w-20',
          toggle: 'bg-accent',
        },
      },
    },
  });
</script>

<Preview>
  <ThemeSwitch />
</Preview>

```

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
gavingolden: I figured we didn't need to expose the details of the component composition via the prop api, although maybe that's easier to understand 🤷‍♂️

techniq:
Deployed as [0.77](https://github.com/techniq/svelte-ux/releases/tag/svelte-ux%400.77.0)

PR #530, opened 2025-01-08
next: migrate to svelte 5

00casanova00:
This PR migrates Svelte UX to Svelte 5. Key changes include:

- Migrated all components to Svelte 5.
- Migrated the documentation website to Svelte 5.
- Updated all relevant code snippets in the documentation.

Note: This PR is a submission for the Svelte Hackathon. Any feedback received will be addressed following the announcement of the winners.

PR #531, opened 2025-01-13
add bind:inputEl to textarea in TextField

### LINKED ISSUE #357: TextField with multiline doesn't get inputEl set
In `TextField`, only the `Input` element has `bind:inputEl`. If it is a multiline `TextField`, `inputEl` will always be null. Unless there are considerations on the differences between `textarea` and `Input` that I am not aware of, adding `bind:this={inputEl}` to the `textarea` seems to work.

hahn-kev:
closes #357 

I didn't test this as I just checked out this repo for this simple change, I'm assuming CI will catch any issues this causes (if any)

### DIFF
--- a/packages/svelte-ux/src/lib/components/TextField.svelte
+++ b/packages/svelte-ux/src/lib/components/TextField.svelte
@@ -367,7 +367,7 @@
                 {max}
                 {step}
                 {actions}
-                bind:inputEl
+                bind:inputEl={null, (el) => (inputEl = el ?? null)}
                 on:input={handleInput}
                 on:focus
                 on:blur
--- a/packages/svelte-ux/src/lib/components/TextField.svelte
+++ b/packages/svelte-ux/src/lib/components/TextField.svelte
@@ -61,7 +61,7 @@
     ? (node) => [autoFocus(node, typeof autofocus === 'object' ? autofocus : undefined)]
     : undefined;
   export let operators: { label: string; value: string }[] | undefined = undefined;
-  export let inputEl: HTMLInputElement | null = null;
+  export let inputEl: HTMLInputElement | HTMLTextAreaElement | null = null;
   export let debounceChange: boolean | number = false;
   export let classes: {
     root?: string;
@@ -327,6 +327,7 @@
                 {disabled}
                 value={inputValue}
                 {autocapitalize}
+                bind:inputEl
                 on:input={handleInput}
                 on:focus
                 on:blur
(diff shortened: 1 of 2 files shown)


packages/svelte-ux/src/lib/components/TextField.svelte
```diff
@@ -366,7 +367,7 @@
                 {max}
                 {step}
                 {actions}
-                bind:inputEl
+                bind:inputEl={null, (el) => (inputEl = el ?? null)}
```
hahn-kev, line 370: this is using function bindings as defined [here](https://svelte.dev/docs/svelte/bind#Function-bindings). The problem was that svelte thought we wanted to provide the value from TextField to Input, and since `inputEl` now may be a TextArea it was not happy with that since there's missing properties. This ensures that data flow is only one way out of the Input and into the TextField variable. I would have put a comment in code but there's not a good way to do that here.

techniq: Thanks @hahn-kev.  As of right now Svelte UX supports Svelte 3-5 (and will be for 1.x, but will be Svelte 5 only).  I plan to release Svelte UX 1.0 VERY soon (maybe later this week) and have a [PR](https://github.com/techniq/svelte-ux/pull/530) that has started the migration to Svelte 5, but there is a lot of regression testing and refinements that need done.

With that said, we won't be able to use Svelte 5 function bindings yet (the docs are running Svelte 5 so it succeeds).

Here's my [rough plan](https://github.com/techniq/svelte-ux/pull/530#issuecomment-2582849134) for Svelte UX 2.x (and others).  Once we start the `next` branch I would love to include this, and plan to have incremental releases.

hahn-kev: Ahh ok, I didn't realize that was a svelte 5 feature. I've changed this to do the same thing but an object to act as a wrapper which lets us define a setter without a getter.

techniq: Nice technique!  Care to run `pnpm format` to fix the prettier lint issue and I'll merge / release.

techniq: Also, could you add an example or a way to validate the change.  I don't have any component tests at the moment, just some unit tests and a lot of examples for visual/manual validation

hahn-kev: fixed the lint issues. FYI maybe this is because I'm on windows, but when I ran format it modified 486 files, in addition to the TextField file. I only checked in that one. That might trip up others though, I think it might be a BOM issue, but it's hard to tell what the changes are.

hahn-kev: Alright there's now also an example, here's a screenshot of it, I figured I should include a toggle between single and multi-line since that's what this PR was mainly for.
image

techniq (approved):

techniq:
I still see some lint issues, but I'll fix them after merging.  Thanks for the PR!

techniq:
Actually the failure was due to a test setup (and unrelated to this PR)

techniq:
Released as [0.77.1](https://github.com/techniq/svelte-ux/releases/tag/svelte-ux%400.77.1).  Thanks again!

PR #541, opened 2025-02-07
Resolve negative counts in pagination

brandonmcconnell:
Resolve a bug causing negative counts in pagination when there are fewer items than the per page count

### DIFF
--- a/.changeset/eleven-spiders-agree.md
+++ b/.changeset/eleven-spiders-agree.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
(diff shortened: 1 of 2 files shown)


techniq (approved):
Thanks @brandonmcconnell.  This should provide a safe guard, although I think updating the [paginationStore](https://github.com/techniq/layerstack/blob/main/packages/svelte-stores/src/lib/paginationStore.ts#L53) should be better handled too.

maybe instead of:
```js
from: Math.min(total, (page - 1) * perPage + 1)
```
it should be:
```js
from: Math.min(total, Math.max(0, (page - 1) * perPage + 1))
```
Regardless, this will protect from the view side.  Thanks!

PR #548, opened 2025-02-13
Replace svelteUx -> layerstack in docs

jostrzol:
I'm not sure if the `colorSpace` argument is still relevant though.


techniq (approved):
Thanks @jostrzol, and `colorSpace` is still valid, but optional (defaults to `hsl`)

PR #590, opened 2025-05-14
apply component setttings classes for Card

lnhua:
There are 4 classes in Card:
```
  export let classes: {
    root?: string;
    header?: ComponentProps<Header>['classes'];
    headerContainer?: string;
    content?: string;
    actions?: string;
  } = {};
```
but only root is applied.


packages/svelte-ux/src/lib/components/Card.svelte
```diff
@@ -48,23 +48,23 @@
   {/if}
 
   {#if title || subheading || $$slots.header}
-    <div class={cls('p-4', classes.headerContainer)}>
+    <div class={cls('p-4', settingsClasses.headerContainer, classes.headerContainer)}>
       <slot name="header">
-        <Header {title} {subheading} classes={classes.header} />
+        <Header {title} {subheading} classes={{ ...settingsClasses.header, ...classes.header }} />
```
techniq, line 53: There is a `clsMerge()` util to properly handle class overrides (for example: `p-2` in `classes.header` should override `p-1` in `settingsClasses.header`).  See `SelectField` / `MultiSelectField` / `ThemeSwitch` for example usage

techniq (approved):
Thanks @lnhua.  I added a quick example to the bottom of the Card docs and also created a changeset (`pnpm changeset`) so this will be picked up in the CHANGELOG and bump the version for publishing.

PR #609, opened 2025-06-16
Updated the info message and ran formatting check

0xmiki:
- removed svelte5 from banner 
- added improved info message
- ran pnpm format


techniq (approved):
Thanks @0xmiki!

PR #629, opened 2025-08-05
feature(NumberStepper): Add slots to passthrough prefix and suffix to the NumberStepper component

MaCleodWalker:


techniq (approved):
Thanks @MaCleodWalker, makes sense.  So looking forward to getting Svelte UX migrated to Svelte 5 snippets which will not require this.

Can you generate a changeset (`pnpm changeset`) for this so the version bumps.  Thanks!

PR #638, opened 2025-08-12
fix(SelectField): keep menu open on reactive updates; close only on selection or outside click

MaCleodWalker:


techniq (approved):
This is great, thanks @MaCleodWalker.

I do feel like there is a use case where you also don't want the selection modified as you are changing the "beforeOptions" filters (for example if you don't click on an option and just click away, the user probably assumes the state was unchanged, especially if the selection wouldn't be filtered by the active filtering(but could have been in the process).

That's going to be challenging to support and also reactively update when the options change AND you want them to update.

Regardless, this is a great change.  Thanks!

PR #671, opened 2026-06-17
Add events to theme components

justlilith:
I noticed that there wasn't a way to tell what theme a user has selected, so this PR adds a new event (carrying a detail of the selected theme) when setting the theme.


techniq:
Hi @justlilith, thanks for the PR.  At the moment Svelte UX has not been ported to Svelte 5 syntax (runes, snippets, etc) but is in the process of as part of a larger effort to consolidate it with LayerStack.  I'll be posting more details soon.

justlilith:
@techniq Thanks; sounds good! In the interim, should I replace the host rune with the deprecated createEventDispatcher so this might be merged? Or just hold off on this PR until the Svelte 5 migration?

techniq:
If you would want to use `createEventDispatcher` ( we are elsewhere in the project) that would be great and get this landed now, as the migration to Svelte 5 will take some time.  Thanks!

packages/svelte-ux/src/lib/components/ThemeSwitch.svelte
```diff
@@ -31,7 +34,7 @@
     otherClasses,
     classes
   )}
-  {...$$restProps}
+  {...props}
```
techniq: We need to switch back to Svelte 4 `$$restProps` syntax as well

package.json
```diff
@@ -23,5 +23,9 @@
     "rimraf": "6.1.3",
     "wrangler": "^4.76.0"
   },
-  "packageManager": "pnpm@10.5.1"
+  "packageManager": "pnpm@10.5.1",
+  "dependencies": {
+    "@layerstack/tailwind": "2.0.0-next.22",
+    "@sveltejs/adapter-cloudflare": "^7.2.8"
+  }
```
techniq: these shouldn't be added here

.changeset/clean-eggs-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'svelte-ux': minor
+---
+
+This release adds events to theme selection.
```
techniq: let's change this to `Add theme selection events`

techniq:
Looks good!  Could you run `pnpm format` to run prettier and make CI happy

justlilith:
@techniq all set, I think 👍

techniq:
Thanks again @justlilith!  Released and available as [2.0.0-next.24](https://github.com/techniq/svelte-ux/releases/tag/svelte-ux%402.0.0-next.24).

FYI: We're currently in the [process](https://github.com/techniq/layerstack/pull/100) of updating LayerStack to use the new docs system we built for LayerChart (and is becoming a new `@layerstack/docs` package), and getting LayerChart 2.0 released (before I go on vacation), then we'll be cranking away on the big overhaul/consolidation of LayerStack/Svelte UX.

I'll post more updates as this works progresses.

PR #673, opened 2026-06-18
Theme events patch (property rename)

justlilith:
This PR renames the detail property in the events dispatched from theme components to theme. This affects usage downstream.

Before:

```ts
on:themeSet={(e) => {
  console.log(e.detail.detail);
}}
```

After:

```ts
on:themeSet={(e) => {
  console.log(e.detail.theme);
}}
```

It removes redundancy and is a little clearer to read/use.

### DIFF
--- a/.changeset/lucky-gifts-glow.md
+++ b/.changeset/lucky-gifts-glow.md
@@ -1,0 +1,5 @@
+---
+'svelte-ux': patch
+---
+
+Replace event.detail.detail with event.detail.theme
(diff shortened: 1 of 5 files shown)


techniq (approved):

techniq:
Good catch!  I can merge and release now, or you can use the [PR package](https://github.com/techniq/svelte-ux/pull/673#issuecomment-4743010666) to test it out and see if there are any other changes you might need.

justlilith:
I'll use the PR package for now; I also need to add something about the events to the svelte-ux guide.

justlilith:
@techniq Okay, I added the examples. I think this PR is ready from my end.

